PR #159, opened 2014-10-17
PR: Implement "Monte Carlo" brute force "RGB" colourspace volume computation.

### LINKED ISSUE #158: Implement "Monte Carlo" brute force "RGB" colourspace volume computation.

KelSolaar:
Closes #158.


colour/algebra/tests/tests_random.py
```diff
@@ -0,0 +1,63 @@
...
+    unit tests methods.
+    """
+
+    def test_random_triplet_generator(self):
+        """
+        Tests :func:`colour.algebra.random.random_triplet_generator`
+        definition.
+        """
+
+        prng = np.random.RandomState(4)
```
MichaelMauderer: I don't think the expected behavior (consistency of the random numbers) here is guaranteed across systems/versions. [This SO question](http://stackoverflow.com/questions/8786084/reproducibility-of-python-pseudo-random-numbers-across-systems-and-versions?rq=1) seem to confirm this too.
I guess as long as it doesn't make problems this is not a big issue, but using a mock random number generator might be the "more correct" solution here to avoid this from breaking in the future.

KelSolaar: Oh yeah that's true. I haven't looked at the tests code in a while. I will add a reference so that we remember if the tests brake in the future.

MichaelMauderer: The reference also needs to go in `colour/volume/tests/tests_rgb.py`

Otherwise everything looks fine to me.

KelSolaar: Yeah, I also added the mention to the definition themselves since the doctests also use `np.random.RandomState`. We should be covered.

PR #172, opened 2015-01-12
PR: Implement various "colour.models" changes related to "ACES VFX Enablement Working Group" work.

KelSolaar:


colour/models/dataset/aces.py
```diff
@@ -0,0 +1,436 @@
...
+
+    Returns
+    -------
+    numeric
+        Companded value.
+    """
+
+    constants = ACES_PROXY_CONSTANTS.get(bit_depth)
+
+    float_2_cv = lambda x: max(constants.CV_min,
```
MichaelMauderer: Might it be clearer to use [`numpy.clip`](http://docs.scipy.org/doc/numpy/reference/generated/numpy.clip.html)?

KelSolaar: Yes most likely! I decided to leave it as is because it is implemented this way in the Technical Bulletin.

MichaelMauderer:
Okay. Well it's not a major thing. 
Everything else looks fine to me.

KelSolaar:
Ok excellent! Thanks!

PR #174, opened 2015-01-21
PR: Add "Simplified sRGB" colourspace.

crowsonkb:
The "Simplified sRGB" colorspace is just sRGB with a transfer function of gamma 2.2 instead of the one defined in the standard. Actual displays tend to implement simplified sRGB instead of the standard's more complicated transfer function. The name "Simplified sRGB" came from Adobe Photoshop's name for this colorspace that more closely matches the operation of real-world devices. I find it useful to compute using it at times, especially when I know I am producing images that will be displayed on devices without color management.


KelSolaar:
Hi Katherine,

Thanks for the PR! My thoughts about it:

- Is there a link online somewhere to a paper or an ICC profile for the colourspace or is it only available in Photoshop as a configuration? We ideally base everything against references (Papers, urls, code, etc...)
- Since it is almost *sRGB*, the code should belong to the `colour.models.dataset.srgb` module.

That being said I'm a bit reluctant to include that colourspace variation since it is just an OECF change but I guess if you have some references it should be fine, my point is that if we let stuff being added to the API without clues where it is coming from, in a few months, years we may not be able to track its origin.

crowsonkb:
That sounds reasonable. I'll try to dig up some references on how the sRGB colorspace was initially constructed to justify this. Also, I had Photoshop export/generate an ICC profile for it which I stuck on my Dropbox just now: https://www.dropbox.com/s/a1ycumese7oskql/Simplified%20sRGB%20IEC61966-2.1.icc?dl=0

...It should probably find a more permanent home but for the purpose of this one comment it's probably OK.

KelSolaar:
I searched a bit and didn't find anything really relevant except for a mention on Bruce Lindbloom website: http://www.brucelindbloom.com/index.html?WorkingSpaceInfo.html

It should be enough.

KelSolaar:
@crowsonkb Katherine, I'll close that PR for now! Cheers!

PR #188, opened 2015-06-30
PR: Implement support for illuminant arguments with Luminance value in various "colour.models" definitions.

### LINKED ISSUE #185: Colour models definitions using an illuminant / whitepoint argument should use its tristimulus values instead of its chromaticity coordinates.
Various definitions using an illuminant argument are right now expecting the chromaticity coordinates of that illuminant as an input: https://github.com/colour-science/colour/blob/0df2ae3f1880d402f8d1f0600c38b34adc5f2a03/colour/models/cie_lab.py#L48 

However in some circumstances where a custom illuminant is needed and the illuminant Luminance is not normalised, it can creates discrepancies with ground truth / expected values.

Most of the definitions end up converting back the chromaticity coordinates to tristimulus values as a first step anyway.

@MichaelMauderer if you have more details, please feel free to add them. I will add a list of the offending objects latter.

KelSolaar:
Closes #185.

### DIFF
--- a/colour/models/cie_xyy.py
+++ b/colour/models/cie_xyy.py
@@ -9,6 +9,8 @@ Defines the *CIE xyY* colourspace transformations:
 
 -   :func:`XYZ_to_xyY`
 -   :func:`xyY_to_XYZ`
+-   :func:`xy_to_xyY`
+-   :func:`xyY_to_xy`
 -   :func:`xy_to_XYZ`
 -   :func:`XYZ_to_xy`
 
@@ -39,6 +41,8 @@ __status__ = 'Production'
 
 __all__ = ['XYZ_to_xyY',
            'xyY_to_XYZ',
+           'xy_to_xyY',
+           'xyY_to_xy',
            'xy_to_XYZ',
            'XYZ_to_xy']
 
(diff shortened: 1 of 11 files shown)


colour/models/cie_xyy.py
```diff
@@ -134,6 +138,112 @@ def xyY_to_XYZ(xyY):
     return XYZ
 
 
+def xy_to_xyY(xy):
+    """
+    Converts from *xy* chromaticity coordinates to *CIE xyY* colourspace.
+
+    `xy` argument with last dimension being equal to 3 will be assumed to be a
```
MichaelMauderer: Should this explicitly mention what happens if this is not already a xyY coordinate? (i.e., Y=1 will be assumed)

KelSolaar: What about something along those lines:

> Converts from *xy* chromaticity coordinates to *CIE xyY* colourspace by extending the array last dimension with :math:`Y` Luminance.

MichaelMauderer: Sounds good!
Maybe also add the actual value used for Y.
Btw, could the Y actually be an optional value of the method with default 1?

KelSolaar: I was actually thinking about that when I wrote the comment, and thought uses could just multiply the appropriate index, but yeah lets do that, it will be much nicer.

KelSolaar: Now that I was adding the argument, I found it a bit weird actually to have that option, in cases we need Luminance to be passed, shouldn't it be through an *xyY* array?

Issue #64, opened 2014-06-11
Implement support for colour blindness.


KelSolaar:
## References
- Brettel, H., Viénot, F., & Mollon, J. D. (1997). Computerized simulation of color appearance for dichromats. Journal of the Optical Society of America. A, Optics, Image Science, and Vision, 14, 2647–2655. doi:10.1364/JOSAA.14.002647

KelSolaar:
[Color.jl](https://github.com/JuliaLang/Color.jl/blob/master/src/algorithms.jl) implements the code from Brettel, H., Viénot, F. and Mollon, J. D. (1997).

PR #192, opened 2015-07-13
PR: Simulation of CVD - Machado (2009)

KelSolaar:
References #64.


colour/blindness/machado2010.py
```diff
@@ -0,0 +1,271 @@
...
+        cmfs = cmfs.clone().interpolate(SpectralShape(steps=1))
+
+    M_n = RGB_to_WSYBRG_matrix(cmfs, primaries)
+    cmfs_a = anomalous_trichromacy_cmfs_Machado2010(cmfs, d_LMS)
+    M_a = RGB_to_WSYBRG_matrix(cmfs_a, primaries)
+
+    return dot_matrix(np.linalg.inv(M_n), M_a)
+
+
+def cvd_matrix_Machado2010(deficiency, severity):
```
KelSolaar: I would like to find a better name for that definition, it basically does almost the same job than `colour.anomalous_trichromacy_matrix_Machado2010 ` definition except that it is using the pre-computed matrices from `colour.blindness.dataset.machado2010` module.

colour/blindness/machado2010.py
```diff
@@ -0,0 +1,271 @@
...
+    alpha = lambda x: (20 - x) / 20
+
+    # Corrected equations as per:
+    # http://www.inf.ufrgs.br/~oliveira/pubs_files/
+    # CVD_Simulation/CVD_Simulation.html#Errata
+    L_a = alpha(d_L) * L + 0.96 * area_L / area_M * (1 - alpha(d_L)) * M
+    M_a = alpha(d_M) * M + 1 / 0.96 * area_M / area_L * (1 - alpha(d_M)) * L
+    # TODO: Check inconsistency with ground truth values, d_S domain seems
+    # to be [5, 59] instead of [0, 20].
+    S_a = cmfs.s_bar.clone().shift(d_S).values
```
KelSolaar: The main issue in our implementation lies probably here:
-  The shift to somehow match the dataset *Tritanomaly* matrices seems to be [5, 59] instead of [0, 20], there is probably something going on with the fact the *Smith & Pokorny 1975 Normal Trichromats* CMFS have 5 nanometers bins.
- Which leads me to the fact that I would like to find a way to offset any spd by an arbitrary offset, right now what I'm doing is interpolating the spectra to 1 nanometer bin and then shifting the indexes of the required nanometers amount but say that you want a 0.01nm shift, it would be quite expensive to interpolate the SPD at 0.01nm.

KelSolaar:
@MichaelMauderer : So looking at the pre-generated matrices and the shift we induce for Tritanomaly, it seems to make sense, e.g. a 59nm shift as suggested by the highest value pre-generated matrix:

``` python
import numpy as np

import colour
from colour.plotting import *

colour_plotting_defaults()

cmfs = colour.LMS_CMFS['Smith & Pokorny 1975 Normal Trichromats']
cmfs_cvd = colour.anomalous_trichromacy_cmfs_Machado2009(cmfs, np.array([0, 0, 59]))
colour.CMFS[cmfs_cvd.name] = cmfs_cvd
multi_cmfs_plot(['Smith & Pokorny 1975 Normal Trichromats', cmfs_cvd.name], legend_location='lower right')
```

image
 
In the light of this comment in the paper, I think we should be fine:

```
Since there are no strong biological explanations yet to justify the causes of tritanopia
and tritanomaly, tritanomaly is simulated based on the shift paradigm only (Equation 4.19) as an approximation to the actual phenomenon and restrain the proposed model from try- ing to model tritanopia. Likewise, simulation of monochromacy (either rod or cone) is not supported by this model. One should note, however, that the conditions covered by the proposed model (i.e., anomalous trichromacy and dichromacy) correspond to approximately 99.96% of all CVD cases (RIGDEN, 1999; SHARPE et al., 1999).
```

I'll add a warning in that regard in our implementation.

What do you think?

KelSolaar:
Thanks @MichaelMauderer!

2.5 years in the making :)

Issue #201, opened 2015-07-18
Fix inconsistencies in support for fractional steps size in "colour.SpectralPowerDistribution" class. 


KelSolaar:
Because of imprecision in floating point representation, fractional steps size could lead to some wavelengths not being represented exactly as the same although they should have been in the perfect world. 

Here is a failing test for example, while interpolating the two following spectral data to 0.1 steps size, one would expect for the spd wavelengths to be all present in the CMFS wavelengths which have a larger shape, but it is not the case:

```
import numpy as np

import colour

spd = colour.SMITS_1999_SPDS['White'].clone().interpolate(colour.SpectralShape(steps=0.1))
cmfs = colour.CMFS['CIE 1931 2 Degree Standard Observer'].clone().interpolate(colour.SpectralShape(steps=0.1))
np.all(np.in1d(spd.wavelengths, cmfs.wavelengths))
```

I found two solutions: 
- One implemented solution is to use a rounded version of the wavelengths.
- One untested solution is to use the `Decimal` module to represent wavelengths (which may leads to a lot of various issues and incompatibilities with the current API)

PR #203, opened 2015-07-19
PR: Fix inconsistencies in support for fractional steps size in "colour.SpectralPowerDistribution" class.

KelSolaar:
References #201.

### DIFF
--- a/colour/colorimetry/spectrum.py
+++ b/colour/colorimetry/spectrum.py
@@ -7,6 +7,7 @@ Spectrum
 
 Defines the classes handling spectral data computation:
 
+-   :class:`SpectralMapping`
 -   :class:`SpectralShape`
 -   :class:`SpectralPowerDistribution`
 -   :class:`TriSpectralPowerDistribution`
@@ -29,6 +30,7 @@ from colour.algebra import (
     SplineInterpolator,
     SpragueInterpolator)
 from colour.utilities import (
+    ArbitraryPrecisionMapping,
     in_array,
     is_iterable,
     is_numeric,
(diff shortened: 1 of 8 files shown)


colour/colorimetry/spectrum.py
```diff
@@ -53,6 +56,88 @@
...
+    a specfic decimals count.
+
+    Parameters
+    ----------
+    data : dict or SpectralMapping, optional
+        Spectral data in a *dict* or *SpectralMapping* as follows:
+        {wavelength :math:`\lambda_{i}`: value,
+        wavelength :math:`\lambda_{i+1}`,
+        ...,
+        wavelength :math:`\lambda_{i+n}`}
```
MichaelMauderer: Should this be 
```
{wavelength :math:`\lambda_{i}`: value_0,
wavelength :math:`\lambda_{i+1}`: value_1,
...,
wavelength :math:`\lambda_{i+n}`: value_n}
```

I'm also not sure if `i` is a good variable name here, as it's supposed to denote a constant float (I assume), instead of a incrementing integer. Maybe just add a description of the domain of `i`.

KelSolaar: I'm actually not even sure it is a good representation at all as the steps are not necessarily uniform either.

colour/colorimetry/spectrum.py
```diff
@@ -53,6 +56,88 @@
...
+    wavelength_decimals : int, optional
+        Decimals count the keys will be rounded at.
```
MichaelMauderer: Maybe also mention behavior for `None`, which seems to be allowed (I'd expect it to disable rounding).

KelSolaar: You can't pass `None` to `round` python function as it raises an exception.

I guess this check is misleading: https://github.com/colour-science/colour/pull/203/files#diff-02aa3d9ac0ba4fa40a15adba94060ef8R135

MichaelMauderer: Ah yeah, that's what lead me to believe it should work, Didn't follow with what `round` does with it.

colour/colorimetry/spectrum.py
MichaelMauderer: Argh and I just misinterpreted the LaTex notation. With `i` being the index and `\lambda_{i}` being the variable it should be fine as is.

colour/colorimetry/spectrum.py
KelSolaar: It is a regular condition I put when doing type checking on properties in general, it gives a chance for the underlying object to be initialised. I could remove the entire check block from here as it is only an alias for `self.key_decimals` which is itself also checking for the value to be an `int`.

PR #250, opened 2016-02-07
PR: Prevent extra line breaks on "Windows" when exporting CSV files.

Ron024:
There is an artifact when files are processed as a ASCII that creates an extra line feed at the end of a line in windows.  If the file is written as a binary this is not an issue.

### DIFF
--- a/colour/io/tabular.py
+++ b/colour/io/tabular.py
@@ -278,7 +278,7 @@ def write_spds_to_csv_file(spds,
                                 'with different shapes to CSV file!'))
 
     wavelengths = tuple(spds.values())[0].wavelengths
-    with open(path, 'w') as csv_file:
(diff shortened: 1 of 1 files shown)


KelSolaar:
Thanks @Ron024,

It seems like writing a binary file breaks on Python 3 (http://stackoverflow.com/a/17725590/931625)

I tried that instead (and reverting the newly introduced `wb` argument to `w`) and it seems to be fine on OS X:

```python
        writer = csv.DictWriter(csv_file,
                                delimiter=str(delimiter),
                                fieldnames=['wavelength'] + fields,
                                lineterminator='\n')
```

Would you mind checking if it works for you?

Cheers,

KelSolaar:
Seems to work!

Thanks :)

PR #269, opened 2016-04-27
PR: Implement support for "Y'CbCr" colourspace.

Nick-Shaw:
First version of a set of functions for converting between RGB and Y'CbCr or Yc'Cbc'Cbr' (the 'constant luminance' variant for ITU-R BT.2020 – Rec.2020).

The functions are designed to be as flexible as possible, so they can take input and return values in various integer and floating point ranges, and can use various preset matrices (or user ones) to do the conversion.

The unit tests are not yet sufficiently comprehensive, but this is put out there for feedback before being taken any further.

### DIFF
--- a/colour/models/rgb/tests/tests_ycbcr.py
+++ b/colour/models/rgb/tests/tests_ycbcr.py
@@ -1,0 +1,115 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+"""
+Defines unit tests for :mod:`colour.models.rgb.ycbcr` module.
+"""
+
+from __future__ import division, unicode_literals
+
+import numpy as np
+import unittest
+
+from colour.models.rgb.ycbcr import rgb_to_YCbCr, YCbCr_to_rgb, rgb_to_YcCbcCrc,
+        YcCbcCrc_to_rgb, RANGE, WEIGHT
+
+__author__ = 'Colour Developers'
+__copyright__ = 'Copyright (C) 2013 - 2016 - Colour Developers'
+__license__ = 'New BSD License - http://opensource.org/licenses/BSD-3-Clause'
+__maintainer__ = 'Colour Developers'
+__email__ = 'colour-science@googlegroups.com'
+__status__ = 'Development'
+
+__all__ = ['Testrgb_to_YCbCr', 'TestYCbCr_to_rgb', 'Testrgb_to_YcCbcCrc', 'TestYcCbcCrc_to_rgb']
+
+
+class Testrgb_to_YCbCr(unittest.TestCase):
+    """
+    Defines :func:`colour.models.rgb.ycbcr.rgb_to_YCbCr` definition unit tests
+    methods.
+    """
+
+    def test_rgb_to_YCbCr(self):
+        """
+        Tests :func:`colour.models.rgb.ycbcr.rgb_to_YCbCr` definition.
+        """
+
+        np.testing.assert_almost_equal(
+            rgb_to_YCbCr(np.array([0.75, 0.75, 0.0])),
+            np.array([ 0.65842092,  0.17204301,  0.53060532]),
+            decimal=7)
+            
+        np.testing.assert_almost_equal(
+            rgb_to_YCbCr(np.array([0.25, 0.5, 0.75]), outRange=RANGE['legal_10_YC_int']),
+            np.array([471, 650, 390]),
+            decimal=7)
+
+
+class TestYCbCr_to_rgb(unittest.TestCase):
+    """
+    Defines :func:`colour.models.rgb.ycbcr.YCbCr_to_rgb` definition unit tests
+    methods.
+    """
+
+    def test_YCbCr_to_rgb(self):
+        """
+        Tests :func:`colour.models.rgb.ycbcr.YCbCr_to_rgb` definition.
+        """
+
+        np.testing.assert_almost_equal(
+            YCbCr_to_rgb(np.array([ 0.65842092,  0.17204301,  0.53060532])),
+            np.array([0.75, 0.75, 0.0]),
+            decimal=7)
+            
+        np.testing.assert_almost_equal(
+            YCbCr_to_rgb(np.array([ 471, 650, 390]), inRange=RANGE['legal_10_YC_int']),
+            np.array([ 0.25018598,  0.49950072,  0.75040741]),
+            decimal=7)
+            
+
+class Testrgb_to_YcCbcCrc(unittest.TestCase):
+    """
+    Defines :func:`colour.models.rgb.ycbcr.rgb_to_YcCbcCrc` definition unit tests
+    methods.
+    """
+
+    def test_rgb_to_YcCbcCrc(self):
+        """
+        Tests :func:`colour.models.rgb.ycbcr.rgb_to_YcCbcCrc` definition.
+        """
+
+        np.testing.assert_almost_equal(
+            rgb_to_YcCbcCrc(np.array([0.123, 0.456, 0.789])),
+            np.array([ 0.59258892,  0.64993099,  0.35269847]),
+            decimal=7)
+            
+        np.testing.assert_almost_equal(
+            rgb_to_YcCbcCrc(np.array([0.18, 0.18, 0.18]), outRange=RANGE['legal_10_YC_int'], is_10_bits_system = True),
+            np.array([422, 512, 512]),
+            decimal=7)
+
+
+class TestYcCbcCrc_to_rgb(unittest.TestCase):
+    """
+    Defines :func:`colour.models.rgb.ycbcr.YCbCr_to_rgb` definition unit tests
+    methods.
+    """
+
+    def test_YcCbcCrc_to_rgb(self):
+        """
+        Tests :func:`colour.models.rgb.ycbcr.YCbCr_to_rgb` definition.
+        """
+
+        np.testing.assert_almost_equal(
+            YcCbcCrc_to_rgb(np.array([1689, 2048, 2048]), inRange=RANGE['legal_12_YC_int'], is_10_bits_system = False),
+            np.array([ 0.18009037,  0.18009037,  0.18009037]),
+            decimal=7)
+            
+        np.testing.assert_almost_equal(
+            YcCbcCrc_to_rgb(np.array([0.678, 0.4, 0.6])),
+            np.array([ 0.69100667,  0.47450469,  0.25583733]),
+            decimal=7)
+
+
+if __name__ == '__main__':
+    unittest.main()
--- a/colour/models/rgb/tests/tests_ycbcr.py
+++ b/colour/models/rgb/tests/tests_ycbcr.py
@@ -10,8 +10,9 @@ from __future__ import division, unicode_literals
 import numpy as np
 import unittest
 
-from colour.models.rgb.ycbcr import rgb_to_YCbCr, YCbCr_to_rgb, rgb_to_YcCbcCrc,
-        YcCbcCrc_to_rgb, RANGE, WEIGHT
+from colour.models.rgb.ycbcr import (rgb_to_YCbCr, YCbCr_to_rgb,
+                                     rgb_to_YcCbcCrc, YcCbcCrc_to_rgb,
+                                     RANGE, WEIGHT)
 
 __author__ = 'Colour Developers'
 __copyright__ = 'Copyright (C) 2013 - 2016 - Colour Developers'
@@ -20,7 +21,10 @@ __maintainer__ = 'Colour Developers'
 __email__ = 'colour-science@googlegroups.com'
 __status__ = 'Development'
 
-__all__ = ['Testrgb_to_YCbCr', 'TestYCbCr_to_rgb', 'Testrgb_to_YcCbcCrc', 'TestYcCbcCrc_to_rgb']
+__all__ = ['Testrgb_to_YCbCr',
+           'TestYCbCr_to_rgb',
+           'Testrgb_to_YcCbcCrc',
+           'TestYcCbcCrc_to_rgb']
 
 
 class Testrgb_to_YCbCr(unittest.TestCase):
@@ -36,12 +40,14 @@ class Testrgb_to_YCbCr(unittest.TestCase):
 
         np.testing.assert_almost_equal(
             rgb_to_YCbCr(np.array([0.75, 0.75, 0.0])),
-            np.array([ 0.65842092,  0.17204301,  0.53060532]),
+            np.array([0.65842092, 0.17204301, 0.53060532]),
             decimal=7)
-            
+
         np.testing.assert_almost_equal(
-            rgb_to_YCbCr(np.array([0.25, 0.5, 0.75]), outRange=RANGE['legal_10_YC_int']),
-            np.array([471, 650, 390]),
+            rgb_to_YCbCr(np.array([0.25, 0.5, 0.75]),
+                         K=WEIGHT['Rec.601'],
+                         outRange=RANGE['legal_10_YC_int']),
+            np.array([461, 662, 382]),
             decimal=7)
 
 
@@ -57,20 +63,21 @@ class TestYCbCr_to_rgb(unittest.TestCase):
         """
 
         np.testing.assert_almost_equal(
-            YCbCr_to_rgb(np.array([ 0.65842092,  0.17204301,  0.53060532])),
+            YCbCr_to_rgb(np.array([0.65842092, 0.17204301, 0.53060532])),
             np.array([0.75, 0.75, 0.0]),
             decimal=7)
-            
+
         np.testing.assert_almost_equal(
-            YCbCr_to_rgb(np.array([ 471, 650, 390]), inRange=RANGE['legal_10_YC_int']),
-            np.array([ 0.25018598,  0.49950072,  0.75040741]),
+            YCbCr_to_rgb(np.array([471, 650, 390]),
+                         inRange=RANGE['legal_10_YC_int']),
+            np.array([0.25018598, 0.49950072, 0.75040741]),
             decimal=7)
-            
+
 
 class Testrgb_to_YcCbcCrc(unittest.TestCase):
     """
-    Defines :func:`colour.models.rgb.ycbcr.rgb_to_YcCbcCrc` definition unit tests
-    methods.
+    Defines :func:`colour.models.rgb.ycbcr.rgb_to_YcCbcCrc` definition unit
+    tests methods.
     """
 
     def test_rgb_to_YcCbcCrc(self):
@@ -80,11 +87,13 @@ class Testrgb_to_YcCbcCrc(unittest.TestCase):
 
         np.testing.assert_almost_equal(
             rgb_to_YcCbcCrc(np.array([0.123, 0.456, 0.789])),
-            np.array([ 0.59258892,  0.64993099,  0.35269847]),
+            np.array([0.59258892, 0.64993099, 0.35269847]),
             decimal=7)
-            
+
         np.testing.assert_almost_equal(
-            rgb_to_YcCbcCrc(np.array([0.18, 0.18, 0.18]), outRange=RANGE['legal_10_YC_int'], is_10_bits_system = True),
+            rgb_to_YcCbcCrc(np.array([0.18, 0.18, 0.18]),
+                            outRange=RANGE['legal_10_YC_int'],
+                            is_10_bits_system=True),
             np.array([422, 512, 512]),
             decimal=7)
 
@@ -101,13 +110,15 @@ class TestYcCbcCrc_to_rgb(unittest.TestCase):
         """
 
         np.testing.assert_almost_equal(
-            YcCbcCrc_to_rgb(np.array([1689, 2048, 2048]), inRange=RANGE['legal_12_YC_int'], is_10_bits_system = False),
-            np.array([ 0.18009037,  0.18009037,  0.18009037]),
+            YcCbcCrc_to_rgb(np.array([1689, 2048, 2048]),
+                            inRange=RANGE['legal_12_YC_int'],
+                            is_10_bits_system=False),
+            np.array([0.18009037, 0.18009037, 0.18009037]),
             decimal=7)
-            
+
         np.testing.assert_almost_equal(
             YcCbcCrc_to_rgb(np.array([0.678, 0.4, 0.6])),
-            np.array([ 0.69100667,  0.47450469,  0.25583733]),
+            np.array([0.69100667, 0.47450469, 0.25583733]),
             decimal=7)
 
 
--- a/colour/models/rgb/tests/tests_ycbcr.py
+++ b/colour/models/rgb/tests/tests_ycbcr.py
@@ -82,9 +82,9 @@ class TestRGB_to_YCbCr(unittest.TestCase):
             YCbCr)
 
         rgb = np.tile(rgb, 4)
-        rgb = np.reshape(rgb, (4,3))
+        rgb = np.reshape(rgb, (4, 3))
         YCbCr = np.tile(YCbCr, 4)
-        YCbCr = np.reshape(YCbCr, (4,3))
+        YCbCr = np.reshape(YCbCr, (4, 3))
         np.testing.assert_almost_equal(
             RGB_to_YCbCr(rgb),
             YCbCr)
@@ -150,7 +150,6 @@ class TestYCbCr_to_RGB(unittest.TestCase):
             np.array([208, 131, 99]),
             decimal=7)
 
-
     def test_n_dimensional_YCbCr_to_RGB(self):
         """
         Tests :func:`colour.models.rgb.ycbcr.YCbCr_to_RGB` definition
@@ -163,9 +162,9 @@ class TestYCbCr_to_RGB(unittest.TestCase):
             rgb)
 
         rgb = np.tile(rgb, 4)
-        rgb = np.reshape(rgb, (4,3))
+        rgb = np.reshape(rgb, (4, 3))
         YCbCr = np.tile(YCbCr, 4)
-        YCbCr = np.reshape(YCbCr, (4,3))
+        YCbCr = np.reshape(YCbCr, (4, 3))
         np.testing.assert_almost_equal(
             YCbCr_to_RGB(YCbCr),
             rgb)
@@ -233,9 +232,9 @@ class TestRGB_to_YcCbcCrc(unittest.TestCase):
             YcCbcCrc)
 
         rgb = np.tile(rgb, 4)
-        rgb = np.reshape(rgb, (4,3))
+        rgb = np.reshape(rgb, (4, 3))
         YcCbcCrc = np.tile(YcCbcCrc, 4)
-        YcCbcCrc = np.reshape(YcCbcCrc, (4,3))
+        YcCbcCrc = np.reshape(YcCbcCrc, (4, 3))
         np.testing.assert_almost_equal(
             RGB_to_YcCbcCrc(rgb),
             YcCbcCrc)
@@ -303,9 +302,9 @@ class TestYcCbcCrc_to_RGB(unittest.TestCase):
             rgb)
 
         rgb = np.tile(rgb, 4)
-        rgb = np.reshape(rgb, (4,3))
+        rgb = np.reshape(rgb, (4, 3))
         YcCbcCrc = np.tile(YcCbcCrc, 4)
-        YcCbcCrc = np.reshape(YcCbcCrc, (4,3))
+        YcCbcCrc = np.reshape(YcCbcCrc, (4, 3))
         np.testing.assert_almost_equal(
             YcCbcCrc_to_RGB(YcCbcCrc),
             rgb)
--- a/colour/models/rgb/tests/tests_ycbcr.py
+++ b/colour/models/rgb/tests/tests_ycbcr.py
@@ -10,9 +10,11 @@ from __future__ import division, unicode_literals
 import numpy as np
 import unittest
 
-from colour.models.rgb.ycbcr import (rgb_to_YCbCr, YCbCr_to_rgb,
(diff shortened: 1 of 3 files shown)


KevinJW:
I'll comment that I'd prefer derivations of the magic numbers for the weighting factors in the constant luminance versions.

Ideally I would also not limit them to Rec 2020, as strictly speaking you can do the same for any of the weightings. e.g ITP/ICTCP (http://wwwen.zte.com.cn/endata/magazine/ztecommunications/2016/1/articles/201603/t20160311_448972.html) does something similar.

Slightly different topic, I wonder about adding weightings for less used spaces like 240M or the old FCC/NTSC.

I'm assuming nobody cares about the slightly off Rec709 values from older 709 revisions that was used in MPEG 2?

Kevin

KevinJW:
Now a question about the ranges, does anybody else not like the way you end up with a bunch of repeated numbers to make the various bit depths?

Have you tried using a function for generating  them from the standard formula? Something like (untested)

def scaled_reference_value(v, bits = 8):
  return v * 2 ** (bits - 8)

It is a shame BT.2087 doesn't provide sample conversions as they would be great for testing. Maybe you could use the values from one of the colour bars documents ARIB (publicly accessible) only has YPbPr values though.

Kevin

Nick-Shaw:
I simply took the numbers for the constant luminance version from the Rec.2020 publication. If somebody can point me to the derivation of those numbers, it would certainly be nice to make their derivation more explicit in the code.

I had not seen the ITP document before. Interesting. I would argue that it is another special case, and if required could be added as separate functions. I don't see a way of generalising the Yc'Cbc'Crc' functions for that as it works differently. ITP still applies a transfer function to all three channels (after transforming to LMS) and then matrixes them. Yc'Cbc'Crc' calculates Yc from linear RGB and then applies the same transfer function to that and R and B before matrixing to produce Cbc' and Crc'.

Which weightings are included in the presets is up for debate. The current ones are the common ones I included for convenience. You do not of course have to use the presets. You can put **K=[0.123, 0.456]** or whatever you like for custom weightings.

Similarly for range, you can put **inRange=[0.0, 1.0], outRange=[0, 255, 0, 256]** or whatever you want. I suppose you could add other options where you could write **bitDepthOut=10, legalOut=true, intOut=true** and the ranges would be calculated based on those parameters. That would replace need for the list of preset ranges, but I would still argue to keep the option of **inRange=[0, 255]** to allow for custom non-standard ranges, and so when using full range you have the option to use 4-1019 or 0-1023.

Yes, it would be great to have a set of known results for testing. I could take the colour bar values and manually put them through the published transforms, I suppose, as a double-check.

KevinJW:
BT.2390 has more on ICtPt plus other HDR goodness.

As to the derivation of the numbers BT.2020 Table 4 shows some of it your numbers are 2X the values of Pb, Nb, Pr, Nr etc. SMPTE RP 177 shows how the luma/luminance numbers fall out of the choice of primaries, having looked for something consise, maybe a comment saying they are 'magic' with reference to the spec might be better! I'll let other chime in on this.

Kevin

Nick-Shaw:
What about removing the preset ranges, and putting something like:

def rgb_to_YCbCr(rgb, K=WEIGHT['Rec.709'], inRange=None, outRange=None,
                 inBitDepth=10, inLegal=False, inInt=False,
                 outBitDepth=10, outLegal=True, outInt=False):
    if (inRange==None):
        inRange=[0, 2**inBitDepth - 1]
        if inLegal:
            inRange[0]=16*2**(inBitDepth - 8)
            inRange[1]=235*2**(inBitDepth - 8)
        if not(inInt):
            inRange[0] /= 2.0**inBitDepth - 1
            inRange[1] /= 2.0**inBitDepth - 1
    if (outRange==None):
        outRange=[0, 2**inBitDepth - 1]
        if outLegal:
            outRange[0]=16*2**(outBitDepth - 8)
            outRange[1]=235*2**(outBitDepth - 8)
        if not(inInt):
            outRange[0] /= 2.0**outBitDepth - 1
            outRange[1] /= 2.0**outBitDepth - 1

With this, the bitDepth values etc are used to calculate the ranges, unless inRange=[x, y] is explicitly given.

Nick-Shaw:
Sorry, should have looked up how to include code in a post. It should be legible now!

Nick-Shaw:
> As to the derivation of the numbers BT.2020 Table 4 shows some of it your numbers are 2X the values of Pb, Nb, Pr, Nr etc. SMPTE RP 177 shows how the luma/luminance numbers fall out of the choice of primaries

I realise that the weightings are simply the middle row of the NPM. What I did not notice, until you pointed it out, was that the divisors are simply those required to normalise the ranges of Cbc' and Crc' to -0.5 to 0.5, i.e. 2x the respective values at full yellow, blue, cyan and red. Thanks.

Nick-Shaw:
I've edited the code in my post a few times, which in hindsight was probably a daft thing to do. I won't change it any more. I will incorporate that code into a new commit tomorrow.

KelSolaar:
Thanks @Nick-Shaw! I'll start taking a look at that a bit tonight and read the various comments you and @KevinJW did.

colour/models/rgb/ycbcr.py
```diff
@@ -0,0 +1,359 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+"""
+Y'CbCr Colour Encoding
+======================
+
+Defines the Y'CbCr encoding (Y'CbCr is not a colour space) transformations:
```
KelSolaar: We tend to write **colour space** as **colourspace** in `colour`, that being said, I have checked in ISO 22028-1, and they use **colour space**. I would just change it for consistency, we can debate about the proper term later :)

colour/models/rgb/ycbcr.py
```diff
@@ -0,0 +1,359 @@
...
+
+-   :func:`rgb_to_YCbCr`
+-   :func:`YCbCr_to_rgb`
+-   :func:`rgb_to_YcCbcCrc`
+-   :func:`YcCbcCrc_to_rgb`
+
+References
+----------
+.. [1]  Wikipedia. YCbCr. Retrieved February 29, 2016, from
```
KelSolaar: I'll need to put all those references in our reference manager for the Bibliography output.

colour/models/rgb/ycbcr.py
```diff
@@ -0,0 +1,359 @@
...
+        from https://www.itu.int/dms_pubrec/itu-r/rec/bt/\
+        R-REC-BT.709-6-201506-I!!PDF-E.pdf
+.. [3]  Recommendation ITU-R BT.2020(08/2012). Retrieved February 29, 2016,
+        from https://www.itu.int/dms_pubrec/itu-r/rec/bt/\
+        R-REC-BT.2020-0-201208-S!!PDF-E.pdf
+"""
+
+import numpy as np
+
+from colour.utilities import tsplit, tstack, CaseInsensitiveMapping
```
KelSolaar: Nitpicking here, but `CaseInsensitiveMapping` should be the first import, before `tsplit`.

colour/models/rgb/ycbcr.py
```diff
@@ -0,0 +1,359 @@
...
+__status__ = 'Development'
+
+__all__ = ['WEIGHT',
+           'RANGE',
+           'rgb_to_YCbCr',
+           'YCbCr_to_rgb',
+           'rgb_to_YcCbcCrc',
+           'YcCbcCrc_to_rgb']
+
+WEIGHT = (CaseInsensitiveMapping(
```
KelSolaar: That module attribute will likely be exported at the API top level, thus it probably need to be prefixed, for example: `YCBCR_WEIGHTS`.

No need for the extra parenthesis: `= (Case`

colour/models/rgb/ycbcr.py
```diff
@@ -0,0 +1,359 @@
...
+        {'Rec.601': [0.299, 0.114],
+         'Rec.709': [0.2126, 0.0722],
```
KelSolaar: 1. `'Rec.709'`, I think we are using a key with a space like `'Rec. 709'` which is more commonly found in the literature, that being said, I'm starting to lean toward using `'BT.709'`, it seems more correct,  the complete title being *Rec. ITU-R BT.709-6*. This imply that I'll probably rename the colourspaces themselves too in the **cctf** branch: https://github.com/colour-science/colour/tree/feature/cctf/colour/models/rgb/conversion_functions
2. `[0.299, 0.114]` could be a tuple (non-mutable): `(0.299, 0.114)`

colour/models/rgb/ycbcr.py
```diff
@@ -0,0 +1,359 @@
...
+         'Rec.2020': [0.2627, 0.0593]}))
+"""
+List of preset luma weightings:
+    'Rec.601' : [0.299, 0.114]
+    'Rec.709' : [0.2126, 0.0722]
+    'Rec.2020' : [0.2627, 0.0593]
+"""
+
+RANGE = (CaseInsensitiveMapping(
```
KelSolaar: Same than for `WEIGHTS`, `RANGE` will be exposed at the top level. `YCBCR_RANGES` for example?

colour/models/rgb/ycbcr.py
```diff
@@ -0,0 +1,359 @@
...
+         'legal_10_YC_int': [64, 940, 64, 960],
+         'legal_8': [16./255, 235./255],
+         'legal_8_YC': [16./255, 235./255, 16./255, 240./255],
+         'legal_8_int': [16, 235],
+         'legal_8_YC_int': [16, 235, 16, 240],
+         'full': [0.0, 1.0],
+         'YPbPr': [0.0, 1.0, -0.5, 0.5]}))
+"""
+List of preset ranges:
+    'legal_12' : [256./4095, 3760./4095]
```
KelSolaar: Cool trick: if you add `from __future__ import division, unicode_literals` above your first import like here for example: https://github.com/colour-science/colour/blob/feature/cctf/colour/models/rgb/conversion_functions/aces.py#L47, then you can change `256./4095` to `256/4095`, no need to specify a floating point separator.

colour/models/rgb/ycbcr.py
```diff
@@ -0,0 +1,359 @@
...
+    'legal_8' : [16./255, 235./255]
+    'legal_8_YC' : [16./255, 235./255, 16./255, 240./255]
+    'legal_8_int' : [16, 235]
+    'legal_8_YC_int' : [16, 235, 16, 240]
+    'full' : [0.0, 1.0]
+    'YPbPr' : [0.0, 1.0, -0.5, 0.5]
+"""
+
+
+def rgb_to_YCbCr(rgb, K=WEIGHT['Rec.709'], inRange=RANGE['full'],
```
KelSolaar: 1. `rgb_to_YCbCr` should be written `RGB_to_YCbCr` for consistency with the API: https://github.com/colour-science/colour/blob/develop/colour/models/rgb/rgb_colourspace.py#L398
2. I don't mind having each argument on a separated line too please!
3. `inRange` --> `in_range`, now something important and which makes the point of using tuples even more relevant: `inRange` is getting bound to a list (mutable). It is something to avoid in Python, except if you intend using the side effects of doing so (like in the case of recursion for example): http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments

colour/models/rgb/ycbcr.py
```diff
@@ -0,0 +1,359 @@
...
+                 outRange=RANGE['legal_10_YC']):
+    """
+    Converts an array of R'G'B' values to the corresponding Y'CbCr values.
+
+    Note: For ITU-R BT.2020 (Rec.2020) the rgb_to_YCbCr function is only
```
KelSolaar: Maybe that note could be put in a dedicated section like that: https://github.com/colour-science/colour/blob/develop/colour/models/rgb/rgb_colourspace.py#L432

colour/models/rgb/ycbcr.py
```diff
@@ -0,0 +1,359 @@
...
+    applicable tothe non-constant luminance implementation.
+    The rgb_to_YcCbcCrc function should be used for the constant luminance case
+    See https://www.itu.int/dms_pubrec/itu-r/rec/bt/\
+    R-REC-BT.2020-0-201208-S!!PDF-E.pdf
+
+    Parameters
+    ----------
+    rgb : array_like
+        Input R'G'B' values. These may be floats or integers.
+    K : array of two values
```
KelSolaar: `K : array of two values`: should be `K : array_like` for compatibility with Numpy docstrings.

colour/models/rgb/ycbcr.py
```diff
@@ -0,0 +1,359 @@
...
+        The luma weighting coefficients of red and blue.
+        Default: WEIGHT['Rec.709']
+    inRange : array of two values (optional)
```
KelSolaar: We usually format things like that: https://github.com/colour-science/colour/blob/develop/colour/colorimetry/lightness.py#L216, and put a list of all the possible values. It is quite a burden to support and maintain but this makes them available in the doc, the default value is the first in the list, the remainder is alphabetically sorted. Now as we may go toward a generic function for that, it is maybe not relevant.

colour/models/rgb/ycbcr.py
```diff
@@ -0,0 +1,359 @@
...
+    -------
+    ndarray
+        Y'CbCr array.
+        These may be floats or integer code values, depending on the parameter
+        values given for outRange.
+
+    Examples
+    --------
+    >>> RGB = np.array([1.0, 1.0, 1.0])
+    >>> rgb_to_YCbCr(RGB)  # doctest: +ELLIPSIS
```
KelSolaar: `# doctest: +ELLIPSIS` is used to prevent doctests precision issues, we usually trim everything after the 7th decimal point, thus `array([ 0.91886608,  0.50048876,  0.50048876])` becomes `array([ 0.9188660...,  0.5004887...,  0.5004887...])`

colour/models/rgb/ycbcr.py
```diff
@@ -0,0 +1,359 @@
...
+
+    Note the use of 256 for the max Cb/Cr value, which is required so that the
+    Cb and Cr output is centred about 128. Using 255 would centre them about
+    127.5 meaning that there was no integer code value to represent neutral
+    colours. This does however create the possibility of output values of 256,
+    which cannot be stored in an 8-bit integer.
+    """
+    Kr = K[0]
+    Kb = K[1]
+    intOut = (isinstance(outRange[0], int) and isinstance(outRange[1], int) and
```
KelSolaar: 1. `intOut` ---> `int_out` or `int_o` (underscore case) :)
2. Could do something along that line maybe: `int_o` = all([isinstance(x, int) for x in out_range])`

colour/models/rgb/ycbcr.py
```diff
@@ -0,0 +1,359 @@
...
+    Cr += (outRange[3] + outRange[2])/2
+    if intOut:
+        Y = np.round(Y).astype(int)
+        Cb = np.round(Cb).astype(int)
+        Cr = np.round(Cr).astype(int)
+    YCbCr = tstack((Y, Cb, Cr))
+    return YCbCr
+
+
+def YCbCr_to_rgb(YCbCr, K=WEIGHT['Rec.709'], inRange=RANGE['legal_10_YC'],
```
KelSolaar: Same comments as `rgb_to_YCbCr` overall.

colour/models/rgb/ycbcr.py
```diff
@@ -0,0 +1,359 @@
...
+    ndarray
+        R'G'B' array.
+        These may be floats or integer code values, depending on the parameter
+        values given for outRange.
+
+    Examples
+    --------
+    >>> YCbCr = np.array([502, 512, 512])
+    >>> YCbCr_to_rgb(YCbCr,
+    ...              inRange=RANGE['legal_10_YC_int'])  # doctest: +ELLIPSIS
```
KelSolaar: 1. Not sure we need the `# doctest: +ELLIPSIS` here.
2. Extreme nitpicking: We usually put the ellipsis statement after the function call opening, and put the arguments on the next lines: https://github.com/colour-science/colour/blob/develop/colour/colorimetry/spectrum.py#L895

colour/models/rgb/ycbcr.py
```diff
@@ -0,0 +1,359 @@
...
+        Yc'Cbc'Crc' array.
+        These may be floats or integer code values, depending on the parameter
+        values given for outRange.
+
+    Examples
+    --------
+    >>> rgb = np.array([0.18, 0.18, 0.18])
+    >>> rgb_to_YcCbcCrc(rgb,
+    ...                 outRange=RANGE['legal_10_YC_int'],
+    ...                 is_10_bits_system = True)  # doctest: +ELLIPSIS
```
KelSolaar: We don't need the # doctest: +ELLIPSIS here.

colour/models/rgb/tests/tests_ycbcr.py
```diff
@@ -0,0 +1,126 @@
...
+"""
+Defines unit tests for :mod:`colour.models.rgb.ycbcr` module.
+"""
+
+from __future__ import division, unicode_literals
+
+import numpy as np
+import unittest
+
+from colour.models.rgb.ycbcr import (rgb_to_YCbCr, YCbCr_to_rgb,
```
KelSolaar: Would it be possible to get one import per line please if they don't fit all on the same one, cheers!

colour/models/rgb/tests/tests_ycbcr.py
```diff
@@ -0,0 +1,126 @@
...
+    Defines :func:`colour.models.rgb.ycbcr.YCbCr_to_rgb` definition unit tests
+    methods.
+    """
+
+    def test_YCbCr_to_rgb(self):
+        """
+        Tests :func:`colour.models.rgb.ycbcr.YCbCr_to_rgb` definition.
+        """
+
+        np.testing.assert_almost_equal(
```
KelSolaar: 1. We try to at least do 3 unit tests but we will likely need more for those guys.
2. We will need to check the support for n-dimensional arrays and NAN too: https://github.com/colour-science/colour/blob/develop/colour/colorimetry/tests/tests_lightness.py#L58

colour/models/rgb/tests/tests_ycbcr.py
```diff
@@ -0,0 +1,126 @@
...
+__email__ = 'colour-science@googlegroups.com'
+__status__ = 'Development'
+
+__all__ = ['Testrgb_to_YCbCr',
+           'TestYCbCr_to_rgb',
+           'Testrgb_to_YcCbcCrc',
+           'TestYcCbcCrc_to_rgb']
+
+
+class Testrgb_to_YCbCr(unittest.TestCase):
```
KelSolaar: `Testrgb_to_YCbCr` to `TestRGB_to_YCbCr` please.

colour/models/rgb/ycbcr.py
```diff
@@ -0,0 +1,359 @@
...
+    Yc, Cbc, Crc = tsplit(YcCbcCrc.astype(float))
+    Yc -= inRange[0]
+    Cbc -= (inRange[3] + inRange[2])/2.0
+    Crc -= (inRange[3] + inRange[2])/2.0
+    Yc *= 1.0/(inRange[1] - inRange[0])
+    Cbc *= 1.0/(inRange[3] - inRange[2])
+    Crc *= 1.0/(inRange[3] - inRange[2])
+    b = np.where(Cbc <= 0.0, Cbc*1.9404 + Yc, Cbc*1.5816 + Yc)
+    r = np.where(Crc <= 0.0, Crc*1.7184 + Yc, Crc*0.9936 + Yc)
+    Yc = REC_2020_COLOURSPACE.EOCF(Yc, is_10_bits_system=is_10_bits_system)
```
KelSolaar: Head-up: This guy is going to change as per the [cctf branch](https://github.com/colour-science/colour/blob/feature/cctf/colour/models/rgb/dataset/rec_2020.py#L101)

KelSolaar:
Just did an initial review, it looks good overall, beside the small formatting / consistency bits the things we will need to take a look at are:

- The generic range computation function.
- The n-dimensional and nan unit tests.

I'll fetch the branch tomorrow morning and play with it.

Now that piece will need to be refactored:

```python
if (inRange==None):
        inRange=[0, 2**inBitDepth - 1]
        if inLegal:
            inRange[0]=16*2**(inBitDepth - 8)
            inRange[1]=235*2**(inBitDepth - 8)
        if not(inInt):
            inRange[0] /= 2.0**inBitDepth - 1
            inRange[1] /= 2.0**inBitDepth - 1
    if (outRange==None):
        outRange=[0, 2**outBitDepth - 1, 0, 2**outBitDepth]
        if outLegal:
            outRange[0]=16*2**(outBitDepth - 8)
            outRange[1]=235*2**(outBitDepth - 8)
            outRange[2]=outRange[0]
            outRange[3]=240*2**(outBitDepth - 8)
        if not(inInt):
            outRange[0] /= 2.0**outBitDepth - 1
            outRange[1] /= 2.0**outBitDepth - 1
            outRange[2] /= 2.0**outBitDepth - 1
            outRange[3] /= 2.0**outBitDepth - 1
            if not(outLegal):
                outRange[2]=-0.5
                outRange[3]=0.5
```

1.  `if (inRange==None)` ---> `if in_range is None`
2. `inRange=[0, 2**inBitDepth - 1]` ---> `in_range = np.array([0, 2 ** in_bit_depth - 1])`, benefit is that you can do a single line for things like that: `in_range /= 2.0 ** in_bit_depth - 1`

> Slightly different topic, I wonder about adding weightings for less used spaces like 240M or the old FCC/NTSC.
> I'm assuming nobody cares about the slightly off Rec709 values from older 709 revisions that was used in MPEG 2?

@KevinJW : Well I don't mind having both of those for reference, the more the better, provided it is sourced :)

colour/models/rgb/ycbcr.py
Nick-Shaw: > ```K : array of two values```: should be ```K : array_like``` for compatibility with Numpy docstrings.

Should that still be referred to as ```array_like``` even though it is just a Python list, not a Numpy array? What about if I change it to a tuple?

colour/models/rgb/ycbcr.py
Nick-Shaw: I admit, I wasn't sure what ```# doctest: ELLIPSIS``` was for, and when it was failing Travis CI I copied and pasted various things blindly from other modules.

colour/models/rgb/ycbcr.py
Nick-Shaw: Should I change it to ```YCBCR_WEIGHTS``` in place for now, and let you move it to the top level if you deem it appropriate?

Nick-Shaw:
Here is a short piece of code I wrote for testing the procedural range generation. The previous one still had errors.

```Python
import numpy as np

in_range=None
out_range=None
in_bits=10
in_legal=False
in_int=False
out_bits=10
out_legal=True
out_int=False
if in_range is None:
    if in_legal:
        in_range=np.array([16, 235])
        in_range *= 2**(in_bits - 8)
    else:
        in_range=np.array([0, 2**in_bits - 1])
    if not(in_int):
        in_range=in_range.astype(float)/(2**in_bits - 1)
if out_range is None:
    if out_legal:
        out_range=np.array([16, 235, 16, 240])
        out_range *= 2**(in_bits - 8)
    else:
        out_range=np.array([0, 2**out_bits - 1, 0, 2**out_bits - 1])
    if not(out_int):
        out_range=out_range.astype(float)/(2**out_bits - 1)
    if (out_int and not(out_legal)):
        out_range[3]=2**out_bits
    if (not(out_int) and not(out_legal)):
        out_range[2]=-0.5
        out_range[3]=0.5

print in_range
print out_range
```
Before I incorporate it into the main code, I wonder if I could canvas opinion on three things:

1. What should the default range for Y'CbCr be? Currently I have it as legal range, float, based on 10-bit, i.e. luma ranged 64/1023 to 940/1023, and chroma 64/1023 to 960/1023.

2. What range should full range integer Y'CbCr be? Currently I have it as per JFIF JPEG spec, i.e. [0, 255, 0, 256]. [0, 255, 0, 255] might seem more logical, but that has no integer value for the chroma of monochrome images. What implementations apart from JPEG use full range Y'CbCr, and what do they do?

3. What range should full range float Y'CbCr be? There is a case for it being [0.0, 1.0, -0.5, 0.5] i.e. that of Y'Ecb'Ecr' in the BT.709 spec. But there might also be a case for keeping the results positive floats, i.e. [0.0, 1.0, 0.0, 1.0] so chroma is centred about 0.5.

KevinJW:
My take...

> What should the default range for Y'CbCr be? Currently I have it as legal range, float, based on 10-bit, i.e. luma range, float, based on 10-bit, i.e. luma ranged 64/1023 to 940/1023, and chroma 64/1023 to 960/1023.

yes by default it should be as per the SDI scaling, I believe they are should be defined from the 8 bit values and scaled rather than the 10 bit ones.

> What range should full range integer Y'CbCr be? Currently I have it as per JFIF JPEG spec, 

yes stick to the spec, I think it is quite documented that there is potential for clipped values in 8 bit JPEG due to the scaling they chose, from the JFIF spec...

"Since values of E'y have a range of 0 to 1.0 and those for E 'Cb and E 'Cr have a range of -0.5 to +0.5, Y, Cb, and Cr must be clamped to 255 when they are maximum value."

So you should stick with the scaling and make it explicit you don't clamp in the documentation.

> What range should full range float Y'CbCr be?

My preference is for [0.0, 1.0, -0.5, 0.5] making it more like what the the analogue encoding would be. The fully positive range only happens when you digitize the signal. Just like you wouldn't clamp [4, 1019] unless you wanted to send the signal on the wire.

I think it is good to document how to achieve things like "the signal on the wire", or "the values to put in a file" but not to limit and certainly by default I prefer unconstrained unless there is real case of breaking the principal of least surprise, i.e. everybody expects a different behaviour.

Kevin

Nick-Shaw:
Thanks @KevinJW 

> yes by default it should be as per the SDI scaling, I believe they are should be defined from the 8 bit values and scaled rather than the 10 bit ones.

Is there a spec as to what float representations of legal integer code values should be? The BT.709 document refers to float values only in full (analogue) range. Legal range is defined only in terms of code values.

I looked at what scaling Nuke uses for the output of the Colorspace node when set to Y'CbCr. Y' certainly seems to be based on 8-bit values, as you say, i.e. Y' for white is 235/255 = 0.921569. But Cb and Cr seem to be centred on 0.5, and effectively based on half code values, i.e. ranged from 15.5/255 to 239.5/255. There is certainly some logic to that, but I've never seen it documented anywhere.

> yes stick to the spec, I think it is quite documented that there is potential for clipped values in 8 bit JPEG due to the scaling they chose, from the JFIF spec...

Yes, I've changed the reference in the example I give to refer to the actual JFIF spec document, rather than the Wikipedia page. That will be in my next commit.

> My preference is for [0.0, 1.0, -0.5, 0.5] making it more like what the the analogue encoding would be.

That makes sense. People can always use their own custom ranges, or just add [0.0, 0.5, 0.5] to the standard output.

Nick-Shaw:
> Slightly different topic, I wonder about adding weightings for less used spaces like 240M or the old FCC/NTSC.
>
> I'm assuming nobody cares about the slightly off Rec709 values from older 709 revisions that was used in MPEG 2?

@KevinJW do you have references for the coefficients for those? I can't find the MPEG-2 one at all, and can only find the SMPTE 240M ones in [http://www.poynton.com/PDFs/coloureq.pdf](http://www.poynton.com/PDFs/coloureq.pdf) which I am a little wary of since at least one of the equations appears to be incorrect (SMPTE-240M G' from Y'PbPr).

colour/models/rgb/ycbcr.py
Nick-Shaw: What is the best approach here? I guess when you change the nomenclature, this function will then fail its unit tests, which will flag up that it need fixing to match the new syntax. Presumably I leave it as is for now?

KelSolaar: Yeah I would leave it as is for now, I'm trying to get the [cctf branch](https://github.com/colour-science/colour/blob/feature/cctf/) wrapped so that you can rebase your work on top of develop when it gets merged. I think I'm not too far on finishing on that one, maybe a PR this weekend.

Nick-Shaw:
I've updated based on most of the comments above, and added the generic range computation functions. I have updated the docstrings to reflect my changes, and added more detail and examples. Do I need to repeat the extra detail I have put into the RGB_to_YCbCr docstring in all the others?

I don't see a case where the range functions would need to be used anywhere except where they are called from within the various YCbCr functions. They are not sub-functions, as they are used within the four separate YCbCr functions.

What therefore is the logic for exposing them at the top level API? In fact is there a way of hiding the functions, so they are accessible only to the four functions which use them?

Sorry that I am showing up how much I am learning this process as I go along!

Nick-Shaw:
@Nick-Shaw wrote:
> What range should full range float Y'CbCr be?

@KevinJW wrote:
> My preference is for [0.0, 1.0, -0.5, 0.5] making it more like what the the analogue encoding would be.

That's what the code currently does.

But I've been thinking about it again. Because Y'CbCr refers to digitized signals, which are normally unsigned integer, and Y'PbPr is the analog equivalent, there may be a better case for the default being [0.0, 1.0, 0.0, 1.0] because the functions are named after Y'CbCr, not Y'PbPr. Nuke's Colorspace node gives output ranged [0.0, 1.0, -0.5, 0.5] when you select YPbPr. I reference how to match Nuke's Colorspace node in the docstring.

Using [0.0, 1.0, 0.0, 1.0] for ```out_int=False, out_legal=False``` would mean that the choice of out_legal would simply turn a full to legal scale on and off on a positive signal. This is perhaps more logical.

KelSolaar:
That's in! Thanks a ton @Nick-Shaw, thanks @KevinJW for the input. If we have to make changes, it will be on v0.3.8 integration branch on my side, and bigger stuff for v0.3.9 :)

Issue #258, opened 2016-03-09
About component conversion / transfer functions (OECF / EOCF) implementation.


KelSolaar:
First thing for the context two official definitions [1]:

> **additive RGB colour space**
> colorimetric colour space having three colour primaries (generally red, green and blue) such that CIE XYZ tristimulus values can be determined from the RGB colour space values by forming a weighted combination of the CIE XYZ tristimulus values for the individual colour primaries, where the weights are proportional to the radiometrically linear colour space values for the corresponding colour primaries
> 
> **NOTE 1** A simple linear matrix transformation can be used to transform between CIE XYZ tristimulus values and the 3×3 radiometrically linear colour space values for an additive RGB colour space.
> 
> **NOTE 2** Additive RGB colour spaces are defined by specifying the CIE chromaticity values for a set of additive RGB primaries and a colour space white point, together with a colour component transfer function.

> **colour component transfer function**
> single variable, monotonic mathematical function applied individually to one or more colour channels of a colour space
> 
> **NOTE 1** Colour component transfer functions are frequently used to account for the nonlinear response of a reference device and/or to improve the visual uniformity of a colour space.
> 
> **NOTE 2** Generally, colour component transfer functions will be nonlinear functions such as a power-law (i.e. “gamma”) function or a logarithmic function. However, in some cases a linear colour component transfer function may be used.

I'm not very happy with the way we have the conversion functions implemented at the moment : https://github.com/colour-science/colour/tree/develop/colour/models/rgb

It is not straightforward to get that right from an implementation and naming convention point of views and I'm touching that on a regular basis in an attempt to find the right implementation. I would be keen on doing it once for good instead of breaking stuff on a regular basis.

Right now the current implementation is roughly as follows:

![image](https://cloud.githubusercontent.com/assets/99779/13649566/3aa9ef56-e6a3-11e5-94f1-fc49a2a4ecdb.png)

What bothers me is the following:

- We don't have a proper *gamma* definition. Right now each RGB colourspace using own define it, code is just an exponentiation so no big deal  but I think it could be part of the `conversion_functions` package or elsewhere if we want to add functionality such as Gamma fitting, etc... later. As a matter of fact I needed to do that recently.
- The RGB colourspaces define their component transfer / conversion functions and occasionally some of them are exported to the generic `log` functions module such as *ACEScc*, *Alexa LOG C*. It was a poor attempt at consolidating camera vendors Log conversion functions in a single place.
[...]

PR #270, opened 2016-05-06
PR: Refactor colour component transfer functions implementation.

KelSolaar:
Reference #258.


colour/models/rgb/dataset/eci_rgb_v2.py
```diff
@@ -88,55 +87,27 @@
 """
 
 
-def _eci_rgb_v2_OECF(value):
+def _scale_domain_0_100_range_0_1(value, callable_):
     """
-    Defines the *ECI RGB v2* colourspace opto-electronic conversion function.
+    Scales the domain of given callable to [0, 100] and its range to [0, 1].
```
MichaelMauderer: Maybe this docstring should further specify the signature of the callable?
Right now it implicitly assumes a single numeric object (or ndarray) as input and output.

MichaelMauderer: Looking at its use it might also work as a decoratot style function?

colour/models/rgb/transfer_functions/bt_709.py
```diff
@@ -0,0 +1,109 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+"""
```
MichaelMauderer: Should these be written out here? With abbreviations in parentheses?

colour/models/rgb/transfer_functions/bt_709.py
```diff
@@ -0,0 +1,109 @@
...
+__status__ = 'Production'
+
+__all__ = ['oetf_BT709',
+           'eotf_BT709']
+
+
+def oetf_BT709(value):
+    """
+    Defines *Recommendation ITU-R BT.709-6* OETF (OECF) opto-electronic
+    transfer function.
```
MichaelMauderer: Should oetf be put after the full term to make the function name clearer? Same for following functions.

colour/models/rgb/transfer_functions/pivoted_log.py
```diff
@@ -0,0 +1,117 @@
...
+
+    Examples
+    --------
+    >>> log_decoding_PivotedLog(0.43499511241446726)  # doctest: +ELLIPSIS
+    0.1...
+    """
+
+    value = np.asarray(value)
+
+    return (10 ** ((value * 1023 - log_reference) *
```
MichaelMauderer: Just double checking because my CS sense is tingling: these are supposed to be 1023, not 1024?

colour/utilities/common.py
```diff
@@ -302,3 +305,48 @@ def is_integer(x):
     """
 
     return abs(x - round(x)) <= INTEGER_THRESHOLD
+
+
+def filter_kwargs(function, **kwargs):
```
MichaelMauderer: Is this a good solution to a specific otherwise unsolvable problem?
Seems like this might eat otherwise potentially useful errors.

colour/models/rgb/transfer_functions/bt_709.py
KelSolaar: Do you mean if we should have the *OETF (OECF) and EOTF (EOCF)* part in the title?

colour/models/rgb/transfer_functions/pivoted_log.py
KelSolaar: Yeah it is disturbing! Reason is as follows: the 10-bit codes values are in range [0-1023], 12-bit in range [0-4095], so the maximum reachable values are respectively 1023 and 4095.

colour/utilities/common.py
KelSolaar: It is not particularly sexy, but basically it will allow removing the `**kwargs` argument here for example: https://github.com/colour-science/colour/blob/develop/colour/difference/delta_e.py#L93 and still have the function callable in that generic one: https://github.com/colour-science/colour/blob/develop/colour/difference/delta_e.py#L403

colour/models/rgb/dataset/eci_rgb_v2.py
KelSolaar: Just updated the docstring on that one to make what it does more explicit. I'm not sure we need a decorator as it is just a one shot private isolated function, should we have the need for it elsewhere, we will definitely move it to the `utilities` package and turn it into a proper decorator.

colour/models/rgb/transfer_functions/bt_709.py
KelSolaar: @MichaelMauderer : Whenever you have time, if you can provide more details on that one please! :)

colour/models/rgb/transfer_functions/pivoted_log.py
MichaelMauderer: Okay, and `value` is allowed to be in [0, 1] instead of [0, 1[.

colour/models/rgb/transfer_functions/bt_709.py
MichaelMauderer: Maybe. Or in the description below.

colour/utilities/common.py
MichaelMauderer: You mean so you can pass the same set of `kwargs` to both of them and not get an error?

colour/models/rgb/transfer_functions/pivoted_log.py
KelSolaar: Yes exactly!

colour/utilities/common.py
KelSolaar: Yes, this is the purpose of that strange beast, it pretty much makes definitions with incompatible signatures, compatible.

colour/models/rgb/transfer_functions/pivoted_log.py
MichaelMauderer: Okay just to make sure, since this has some additional consequences: is this relationship between `value` and the result externally specified like this?

Issue #278, opened 2016-08-22
Implement support for "Canon Log 2" and "Canon Log 3" log encoding curves.


KelSolaar:
References
-----------
- https://github.com/cameramanben/LUTCalc/commit/5b39ac5a94728fd120dbe54a1257c77fc4e0011e
- https://drive.google.com/open?id=0B_IQZQdc4Vy8ZGYyY29pMEVwZU0

PR #279, opened 2016-08-23
PR: Implement support for "Canon Log 2" and "Canon Log 3" log encoding curves.

KelSolaar:
References #278.

### DIFF
--- a/colour/models/rgb/transfer_functions/canon_log.py
+++ b/colour/models/rgb/transfer_functions/canon_log.py
@@ -2,13 +2,17 @@
 # -*- coding: utf-8 -*-
 
 """
-Canon Log Encoding
-==================
+Canon Log Encodings
+===================
 
-Defines the *Canon Log* encoding:
+Defines the *Canon Log* encodings:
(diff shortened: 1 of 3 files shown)


colour/models/rgb/transfer_functions/canon_log.py
```diff
@@ -21,11 +25,15 @@
 .. [1]  Thorpe, L. (2012). CANON-LOG TRANSFER CHARACTERISTIC. Retrieved
         from http://downloads.canon.com/CDLC/\
 Canon-Log_Transfer_Characteristic_6-20-2012.pdf
+.. [2]  Canon. (n.d.). EOS C300 Mark II. Retrieved August 23, 2016, from
```
MichaelMauderer: Which document/section are you referencing here exactly?

KelSolaar: I have updated the documentation to provide more details about the origin of the file. It is actually a download (I made a copy referenced in the issue).

Issue #280, opened 2016-08-23
Implement support for "RedWideGamut RGB" colourspace and "Log3G10" log encoding curve.


KelSolaar:
References
-----------
- http://www.reduser.net/forum/showthread.php?148272-RedWideGamutRGB-discuss!
- https://github.com/cameramanben/LUTCalc/commit/8d2f0433072dcb1652754cfac9fad9e24639331e

PR #283, opened 2016-08-24
PR: Implement support for "RedWideGamut RGB" colourspace and "Log3G10" log encoding curve.

Nick-Shaw:
References #280 

Unit tests for Log3G10 still need to be added.


colour/models/rgb/transfer_functions/red_log.py
```diff
@@ -212,16 +220,21 @@ def log_encoding_Log3G10(x):
...
+    >>> log_encoding_Log3G10(0.18, legacy_curve=True)  # doctest: +ELLIPSIS
     0.3333336...
+    >>> log_encoding_Log3G10(0.0)  # doctest: +ELLIPSIS
+    0.0915514...
     """
 
     x = np.asarray(x)
 
-    return np.sign(x) * 0.222497 * np.log10((np.abs(x) * 169.379333) + 1)
+    if legacy_curve:
```
KelSolaar: Wondering if we should have a well defined `if ... else ...` block here. I'm nitpicking really :)

colour/models/rgb/transfer_functions/red_log.py
```diff
@@ -238,14 +253,19 @@ def log_decoding_Log3G10(y):
...
-    >>> log_decoding_Log3G10(1.0 / 3)  # doctest: +ELLIPSIS
+    >>> log_decoding_Log3G10(1.0 / 3, legacy_curve=True)  # doctest: +ELLIPSIS
     0.1799994...
+    >>> log_decoding_Log3G10(1.0)  # doctest: +ELLIPSIS
+    184.3223476...
     """
 
     y = np.asarray(y)
 
+    if legacy_curve:
```
KelSolaar: Same here for the `if ... else ...` thing.

colour/models/rgb/transfer_functions/tests/tests_red_log.py
```diff
@@ -340,22 +340,22 @@ def test_log_encoding_Log3G10(self):
         """
 
         self.assertAlmostEqual(
-            log_encoding_Log3G10(0.0),
+            log_encoding_Log3G10(0.0, legacy_curve=True),
```
KelSolaar: Should we have the full tests now for the new curve?

Nick-Shaw: To cover all the possibilities we should have tests for positive and negative values for both versions of the curve, and also nan/inf tests for both versions.

Issue #282, opened 2016-08-24
"sRGB" and "Rec. 709" colourspaces have different normalised primary matrices.


KelSolaar:
As of *Colour 0.3.8* and the various changes in the RGB colourspaces implementation, a defect was introduced:

*sRGB* colourspace is using *IEC 61966-2-1:1999* normalised primary matrix definition:

```python
sRGB_TO_XYZ_MATRIX = np.array(
    [[0.4124, 0.3576, 0.1805],
     [0.2126, 0.7152, 0.0722],
     [0.0193, 0.1192, 0.9505]])
```

while *Rec. 709* colourspace computes it directly:

```python
REC_709_TO_XYZ_MATRIX = normalised_primary_matrix(
    REC_709_PRIMARIES, REC_709_WHITEPOINT)
```
 
This creates discrepancies in the related computations.

The tricky aspect is that both approaches are valid: for example matching V-Gamut whitepaper computations requires the directly computed normalised primary matrix while matching official sRGB values requires using  *IEC 61966-2-1:1999* normalised primary matrix definition.

We need to find an elegant solution to that problem.

Nick-Shaw:
At least in this instance the _IEC 61966-2-1:1999_ matrix is just the computed NPM rounded to four decimal places. It's more complex where the rounding happens earlier in a process, making subsequent values more inaccurate.

I guess that is what is happening in the V-Gamut white paper. The ACES matrix is given to six decimal places, but calculated using an NPM which has been previously rounded to four decimal places. Correct?

Scary how much of a discrepancy that can create.

KelSolaar:
The V-Gamut white paper defines *V-Gamut* to *Rec. 709* as follows:

```
1.806576 -0.695697 -0.110879
-0.170090 1.305955 -0.135865
-0.025206 -0.154468 1.179674
```

Colour *V-Gamut* to *Rec. 709* yields:

```
1.8065736	-0.6956981	-0.1108786
-0.1700890	1.3059548	-0.1358648
-0.0252057	-0.1544678	1.1796737
```

and Colour *V-Gamut* to *sRGB* yields:

```
1.8063853	-0.6956147	-0.1109453
-0.1699311	1.3058387	-0.1358616
-0.0251630	-0.1544899	1.1797117
```

I'm thinking about implementing a switch attribute in the `colour.RGB_Colourspace` class that will force the computation of the NPM instead of using pre-defined values.

KelSolaar:
[...]

PR #289, opened 2016-09-11
PR: Implement support for initialisation time and derived transformation matrices in "colour.RGB_Colourspace" class.

KelSolaar:
References #282.

### DIFF
--- a/colour/models/rgb/rgb_colourspace.py
+++ b/colour/models/rgb/rgb_colourspace.py
@@ -773,9 +773,9 @@ def RGB_to_RGB_matrix(input_colourspace,
     >>> RGB_to_RGB_matrix(
     ...     sRGB_COLOURSPACE,
     ...     PROPHOTO_RGB_COLOURSPACE)  # doctest: +ELLIPSIS
-    array([[ 0.5287748...,  0.3340201...,  0.1373909...],
-           [ 0.0975583...,  0.8789770...,  0.0233924...],
-           [ 0.0163594...,  0.1066096...,  0.8772254...]])
+    array([[ 0.5288241...,  0.3340609...,  0.1373616...],
+           [ 0.0975294...,  0.8790074...,  0.0233981...],
+           [ 0.0163599...,  0.1066124...,  0.8772485...]])
     """
 
     cat = chromatic_adaptation_matrix_VonKries(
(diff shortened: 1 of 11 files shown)


colour/models/rgb/rgb_colourspace.py
```diff
@@ -345,6 +490,100 @@ def decoding_cctf(self, value):
...
+        """
+        Setter for **self._use_derived_RGB_to_XYZ_matrix** private attribute.
+
+        Parameters
+        ----------
+        value : bool
+            Attribute value.
+        """
+
+        assert isinstance(value, bool), (
```
MichaelMauderer: Does this really have to be a bool, or is it fine if we can extract a truth value from `value` using the `bool` function?

MichaelMauderer: It might be more flexible to just test whether the object supports truthiness by doing something like
 ```
try: 
    bool(value)
except ValueError:
    # do somethign
```

Comment (KelSolaar) [inline colour/models/rgb/rgb_colourspace.py]: Thought I had replied to this, may not have pushed the *Add single comment* button.
One issue with this test is that you can pass something like `bool('False')`, `bool('0')`, `bool('Off')`, etc... and still get `True`.

Overall I think I want to rework all those properties checks in the API, *Bokeh* seems to be doing something smart which might be clean, I need to investigate that.

Comment (MichaelMauderer) [inline colour/models/rgb/rgb_colourspace.py]: True, but that is somewhat the point. All of these have a valid truth value that can be used in most cases. Insisting on `True` and `False` might limit some use cases or might preclude some third party libraries that are using other objects to indicate truthiness.

Comment (KelSolaar) [inline colour/models/rgb/rgb_colourspace.py]: Yes agreed, it was just underlining the weird behaviour of `bool` in case of passing a string or anything to it.

This means that we should also assign `bool(value)` instead of `value` itself here:
`self._use_derived_RGB_to_XYZ_matrix = value` otherwise we might have unexpected and surprising behaviour:

```python
>>> bool({'False': 0})
True
>>> {'False': 0} is True
False
```

MichaelMauderer: Not sure that should be considered unexpected in Python, as using the truth value of [strings ](http://stackoverflow.com/questions/9573244/most-elegant-way-to-check-if-the-string-is-empty-in-python) and [containers ](http://stackoverflow.com/questions/53513/best-way-to-check-if-a-list-is-empty) is quite idiomatic. This also precludes one element numpy arrays or the (non-boolean) results of `and`/`or`.

Replacing it could be done, but I'm not sure the behaviour you describe is more unexpected than the value assigned to some variable to be exchanged for another one.

KelSolaar: It is probably part of a bigger discussion on what we allow users to do in general in our properties, so far the behaviour in the API has been to allow for less than more.

In that context, I'm a bit uncomfortable at the idea that one could stuff literally any object in the underlying attribute.

Issue #281, opened 2016-08-24
Implement support for negative number handling in "colour.gamma" definition.


KelSolaar:
@Nick-Shaw is suggesting we add some more arguments to `colour.gamma` definition:

```
 Is there a case for adding kwargs defining handling of negative input, emulating other implementations? I could imagine four possible options:

- `negative='error'` (current behaviour)
- `negative='mirror'` (Resolve behaviour)
- `negative='passthrough'` (Nuke behaviour)
- `negative='clamp'`
```

KevinJW:
> @Nick-Shaw is suggesting we add some more arguments to colour.gamma definition:
>
>  Is there a case for adding kwargs defining handling of negative input, emulating other implementations? I could imagine four possible options, `negative='error'` (current behaviour) `negative='mirror'` (Resolve behaviour) `negative='passthrough'` (Nuke behaviour) and `negative='clamp'`

passthrough sounds like the wrong name for negative handling, it
describes how the function does what it does rather than the result,
though I'm not that fussed for such simple behaviour. I'm also not
100% on the parameter name. I do support the general idea though.

Kevin

Nick-Shaw:
I didn't spent a lot of time thinking what the parameter or options should be called. I agree they are not great. But I couldn't think of obvious ones. I just wanted to suggest the concept.

KelSolaar:
@Nick-Shaw, @KevinJW: I was going through the opened issues and saw this one, I can implement it just need to know if you have any fancy names for the options?

Nick-Shaw:
Afraid I can't think of anything other than my original suggestions. @KevinJW do you have any thoughts? Particularly on the one where you weren't keen on my suggestions.

KelSolaar:
*error* somehow imply that the function will fail which is not the case as it generates *nan* thus what about *indeterminate* or simply *nan*?

*passthrough* is trickier :)

KelSolaar:
I have updated `colour.gamma_function` definition signature with a new `negative_number_handling` argument with the following options:

-   *Indeterminate*: The behaviour will be indeterminate and
    definition return value might contain *nans*.
-   *Mirror*: The definition return value will be mirrored around
    abscissa and ordinate axis, i.e. Blackmagic Design: Davinci Resolve
    behaviour.
-   *Preserve*: The definition will preserve any negative number in
    `a`, i.e. The Foundry Nuke behaviour.
-   *Clamp*: The definition will clamp any negative number in `a` to 0.

PR #297, opened 2016-12-03
PR: Implement support for negative number handling in "colour.gamma" definition.

KelSolaar:
References #281.

### DIFF
--- a/colour/models/rgb/transfer_functions/gamma.py
+++ b/colour/models/rgb/transfer_functions/gamma.py
@@ -21,6 +21,8 @@ from __future__ import division, unicode_literals
 
 import numpy as np
 
+from colour.utilities import as_numeric
+
 __author__ = 'Colour Developers'
 __copyright__ = 'Copyright (C) 2013-2016 - Colour Developers'
 __license__ = 'New BSD License - http://opensource.org/licenses/BSD-3-Clause'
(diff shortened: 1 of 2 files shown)


colour/models/rgb/transfer_functions/gamma.py
```diff
@@ -51,8 +68,27 @@ def gamma_function(a, exponent=1.0):
...
-    return a ** exponent
+    negative_number_handling = negative_number_handling.lower()
+    if negative_number_handling == 'indeterminate':
+        return as_numeric(a ** exponent)
+    elif negative_number_handling == 'mirror':
+        return as_numeric(np.sign(a) * np.abs(a) ** exponent)
+    elif negative_number_handling == 'preserve':
+        return as_numeric(np.where(a < 0, a, a ** exponent))
+    elif negative_number_handling == 'clamp':
+        return as_numeric(np.where(a < 0, 0, a ** exponent))
```
MichaelMauderer: Should we throw an exception in the final else that indicates that the `negative_number_handling` had an invalid value?

KelSolaar: Yeah definitely!

`raise ValueError('Invalid negative number handling value provided!')` or something along those lines.

MichaelMauderer: Yes! Probably might also be good to include the actual value in the message.

Issue #296, opened 2016-11-21
Implement support for CCT to chromaticity coordinates computation from "Krystek (1985)".


KelSolaar:
References
-----------

- Krystek, M. (1985). An algorithm to calculate correlated colour temperature. Color Research & Application, 10(1), 38–40. doi:10.1002/col.5080100109

PR #298, opened 2016-12-07
PR: Implement support for CCT to chromaticity coordinates computation from "Krystek (1985)".

KelSolaar:
References #296.

As I just bought the Cloud version of the paper (and thus cannot save it), here are the relevant equations:

image

### DIFF
--- a/colour/examples/temperature/examples_cct.py
+++ b/colour/examples/temperature/examples_cct.py
@@ -41,7 +41,7 @@ message_box(('Converting to "CIE UCS" colourspace "uv" chromaticity '
              '"Ohno (2013)" method:\n'
              '\n\t({0}, {1})'.format(CCT, D_uv)))
 print(colour.CCT_to_uv_Ohno2013(CCT, D_uv, cmfs=cmfs))
-print(colour.CCT_to_uv(CCT, D_uv, cmfs=cmfs))
+print(colour.CCT_to_uv(CCT, D_uv=D_uv, cmfs=cmfs))
 
 print('\n')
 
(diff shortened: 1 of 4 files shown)


KelSolaar:
Hi, @OlliNiemitalo,

Our implementation is correct AFAIK and after quickly double-checking.

Here is the reference:

image

Note that the results are not excessively different from other methods:

```python
>>> from colour.temperature import *
>>> CCT_to_uv_Krystek1985(6504)
array([ 0.18376718,  0.30934735])
>>> CCT_to_uv_Ohno2013([6504, 0])
array([ 0.20042808,  0.31033343])
```

I would not expect them to be the same anyway. Also:

```python
>>> 1.54118254 * 10e-4
0.0015411825400000001
````

is obviously very different from:

```python
>>> 1.54118254e-4
0.000154118254
```

Cheers,

Thomas

OlliNiemitalo:
I also had to triple-check to spot the bug as it is not easy to see. Copying verbatim from the paper:
```
>>> 1.54118254 * 10**-4
0.000154118254
```
That's identical with 1.54118254e-4. Using the corrected implementation also the results match the other methods much better:
```
>>> import numpy as np
>>> T = 6504
>>> u = ((0.860117757 + 1.54118254e-4 * T + 1.28641212e-7 * T ** 2) / (1 + 8.42420235e-4 * T + 7.08145163e-7 * T ** 2))
>>> v = ((0.317398726 + 4.22806245e-5 * T + 4.20481691e-8 * T ** 2) / (1 - 2.89741816e-5 * T + 1.61456053e-7 * T ** 2))
>>> np.array([u, v])
array([0.20047404, 0.31029562])
```

KelSolaar:
Ha! I see now :) Very well spotted!

Do you mind opening a PR? please keep the `10**-4` form though if you do.

OlliNiemitalo:
I'll try! (It'll be a first for me.)

KelSolaar:
Awesome, I can help you and we have a contributing guide here that should give you some info: https://www.colour-science.org/contributing/

Issue #255, opened 2016-02-28
Add various camera spectral sensitivities.


KelSolaar:
## References
- https://spectralestimation.wordpress.com/data/

As part of #192, it would be great to also add various camera spectral responses.

scooperly:
This seems like a good point to bring up to Alex Forsythe or Scott Dyer on the ACES team, since they had to acquire spectral data to produce IDTs. I'm not so sure how the companies would feel about releasing the data publicly, but worth a shot :)

Also there have been a number of studies that required measuring spec. sens. for consumer cameras like ( https://www.cis.rit.edu/people/faculty/jwgu/research/camspec/camspec.pdf ) Data could be available from these sources as well.

KelSolaar:
That would be great indeed! A long term goal would be to implement support for their characterisation too, this is also somehow related to #63.

Jinyu Fang (@fangjy88) had planned to contribute some related code, I'm not sure what is the latest on that one.

Nick-Shaw:
I'm not sure that the ACES team produced any of the IDTs themselves. I think they deliberately left it up to the camera manufacturers to create them. Indeed the IDT folder in the ACES repository is called "vendorSupplied", which would support this.

KelSolaar:
@Nick-Shaw : Yeah but they may have access to some spectral data, Jim Houston kindly gave me the [Kinoton spd](https://github.com/colour-science/colour/blob/develop/colour/colorimetry/dataset/light_sources/spds.py) for example. I'm summoning @alexfry here :)

PR #299, opened 2016-12-08
PR: Add various camera spectral sensitivities.

KelSolaar:
References #255.


colour/characterisation/cameras.py
```diff
@@ -0,0 +1,151 @@
...
+        """
+        Setter for **self.x** attribute.
+
+        Parameters
+        ----------
+        value : object
+            Attribute value.
+        """
+
+        raise AttributeError('"{0}" attribute is read only!'.format('red'))
```
KelSolaar: I'm not sure to understand, the setter is above.

```
    @property
    def red(self):
        """
        Property for **self.x** attribute.

        Returns
        -------
        SpectralPowerDistribution
            self.x

        Warning
        -------
        :attr:`RGB_SpectralSensitivities.red` is read only.
        """

        return self.x

    @red.setter
    def red(self, value):
        """
        Setter for **self.x** attribute.

        Parameters
        ----------
        value : object
            Attribute value.
        """

        raise AttributeError('"{0}" attribute is read only!'.format('red'))
```

MichaelMauderer: Maybe I'm confused, but doesn't the `@property` only create the getter? If no setter is specified trying to set the property you should see an `AttributeError: can't set attribute`.

KelSolaar: The setter is defined as per: `@red.setter` and will raise the read only exception: `"red" attribute is read only!`

I suspect I'm missing something here :)

MichaelMauderer:
I think I'm just saying you'd also get an attribute error without implementing the setters explicitly.

Not necessarily anything wrong with doing it, but might not be necessary.

________________________________

KelSolaar:
Ohh right I see! Yeah if it helps trimming the fat we could go for it. That *read only* stuff is re-occurring often through the API. We can probably address that in a later overall pass?

MichaelMauderer:
Ah, hand;t noticed it anywhere else before. But sure, lets look at this some other time then.
Everything else seems fine.

Issue #302, opened 2016-12-22
Increase the velocity of the calculations.


iCagarro:
Hello,

I have been working with this library and I wonder if it is possible to increase the velocity of the calculations of the CRI, CQS or CCT, for example. I am using a certified software (LightMtrX) and it can perform a very fast calculation of the CRI, CCT, luminous flux or color coordinates, giving certified measures. However, this library is very useful for other routines, so it would be a very nice issue to solve.

Thank you.

KelSolaar:
@iCagarro : The spectral code path is a bit computation heavy, it can be optimised but I wouldn't expect massive speedup as the codebase is using Python. Which CCT algorithm were you using? Ohno (2013)?

iCagarro:
@KelSolaar I'm using the Robertson1968 because it is the CCT algorithm used in the calculation of the CRI. Maybe reducing the size of the matrix (reducing, in this way, the wavelength resolution) could help to reduce also the computation time.

KelSolaar:
Ah yeah Robertson (1968) is not vectorised because not trivial to do so and has a slow _for_ loop. I might take a look at it though as it is planned somehow. That being said I suspect that the spectral code path is the culprit here.

KelSolaar:
Confirming that the spectral code path is what slows things down, did a bit of optimisation (±30% faster) as per 6503cbf802109477dbcb2e86a92b60d2fcd81adf but it is obviously still very slow. I can probably shave some milliseconds here and there but nothing huge without a massive overhaul of the spectral code.

KelSolaar:
Code in 2616fc88518746b55fa37be61fd93bd0aa4ab498 is 2 times faster than in https://github.com/colour-science/colour/tree/develop. I'll continue and report as I go.

KelSolaar:
Code in aecfdc16aca8e857f387c205ef9c729f07c5f5ad is now ±5.8 times faster compared to https://github.com/colour-science/colour/tree/develop.

KelSolaar:
`colour.colour_rendering_index` is now ±10.5 times / one order of magnitude faster!

KelSolaar:
Summary as of d86d25542b71fcc2ae13de072911f3179193cd56:

| Object                              | Speed Increase (±) |
|-------------------------------------|--------------------|
| colour.SpectralPowerDistribution    |          6         |
| colour.TriSpectralPowerDistribution |          6         |
| colour.colour_rendering_index       |        10.5        |
[...]

PR #304, opened 2016-12-27
PR: Spectral Optimization

KelSolaar:
References #302.


colour/colorimetry/spectrum.py
```diff
@@ -839,7 +835,7 @@ def items(self):
             Spectral power distribution data generator.
         """
 
-        return self.__iter__()
+        return np.array(list(self.__iter__()))
```
MichaelMauderer: Could the intermediary list be avoided by using [`np.fromiter`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.fromiter.html)?

KelSolaar: Very good point, I'm sure it would work.

KelSolaar: Turns out that I didn't manage to get `np.fromiter` to interpret correctly the input:

`return np.fromiter(self.__iter__(), dtype=np.float_)`

```
  File "/Users/kelsolaar/Documents/Development/colour-science/colour/colour/colorimetry/spectrum.py", line 838, in items
    return np.fromiter(self.__iter__(), dtype=np.float_)
ValueError: setting an array element with a sequence.
```

Seems like `np.fromiter` can only construct 1D arrays: http://stackoverflow.com/a/34020332/931625

Issue #313, opened 2017-03-22
Handle interpolation code changes introduced with "Scipy" 0.19.0.


KelSolaar:
[Scipy 0.19.0](https://github.com/scipy/scipy/blob/master/doc/release/0.19.0-notes.rst#other-changes) introduced backward compatible breaking changes in `scipy.interpolate.interp1d` definition, as a result [our builds are breaking](https://travis-ci.org/colour-science/colour/builds/213341301).

We need to take a look on how to handle that properly.

MichaelMauderer:
Would it be acceptable to reduce the significant digits we use for testing? It seems that this mostly affects the last one or two we look at. Not nice but it might also protect us in the future from similar changes.

Not sure that solves issues with doc strings, though, which we probably still would have to modify all u hand at this point.

MichaelMauderer:
Also an interim solution: what about specifying the version that does not yet include this change for our tests?

KelSolaar:
> Not sure that solves issues with doc strings, though, which we probably still would have to modify all u hand at this point.

Yeah it is the big issue at play here, we might have to update a lot of stuff, I'm not against it though but it is not very elegant, maybe we can annotate the code areas we lower the precision and go back to them with full precision once we elect having Scipy 0.19.0 as the minimal version?

> what about specifying the version that does not yet include this change for our tests?

It might be only very temporary, any update of conda will likely break Colour again.

KelSolaar:
I went for the interim solution, this is going to be annoying but at least we have some headroom to continue working now.

PR #314, opened 2017-03-23
Change "colour.eotf_BT1886" and "colour.oetf_BT1886" definitions default arguments.

### LINKED ISSUE #312: Change "L_B" and "L_W" default argument values for "BT.1886" eotf.
Current default values are 64 and 940, i.e. 10-bit black and white code values. L_B and L_W should be screen luminance in cd/m^2 if the function is to return absolute luminance, or e.g. 0 and 1 to return normalised values.

KelSolaar:
Closes #312.

### DIFF
--- a/colour/models/rgb/transfer_functions/bt_1886.py
+++ b/colour/models/rgb/transfer_functions/bt_1886.py
@@ -41,7 +41,7 @@ __all__ = ['oetf_BT1886',
            'eotf_BT1886']
 
 
-def oetf_BT1886(L, L_B=64, L_W=940):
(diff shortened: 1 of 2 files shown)


KelSolaar:
@MichaelMauderer, @Nick-Shaw: The tests are failing because of #313, so you can dismiss the build failing.

KelSolaar:
@Nick-Shaw : Yeah it will makes things a bit easier too, less corner cases to handle.

KelSolaar:
@MichaelMauderer : I don't think we need them as they are not mentioned anywhere else in the transfer functions, now that being said, when we add metadata support for them it will come for free.

PR #319, opened 2017-04-19
Implement support for "Prismatic" colourspace.

### LINKED ISSUE #223: Implement support for "Prismatic" colourspace.
## References
- http://www.cs.utah.edu/~shirley/papers/prismatic.pdf
- http://psgraphics.blogspot.co.nz/2015/10/prismatic-color-model.html

KelSolaar:
Closes #223.

### DIFF
--- a/colour/models/rgb/prismatic.py
+++ b/colour/models/rgb/prismatic.py
@@ -1,0 +1,112 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+"""
+Prismatic Colourspace
+=====================
+
+Defines the *Prismatic* colourspace transformations:
+
+-   :func:`RGB_to_Prismatic`
+-   :func:`Prismatic_to_RGB`
+
+See Also
+--------
+`Prismatic Colourspace Jupyter Notebook
+<http://nbviewer.jupyter.org/github/colour-science/colour-notebooks/\
+blob/master/notebooks/models/prismatic.ipynb>`_
+
+References
+----------
+.. [1]  Shirley, P., & Hart, D. (2015). The prismatic color space for rgb
+        computations, 2–7.
+"""
+
+from __future__ import division, unicode_literals
+
+import numpy as np
+
+from colour.utilities import tsplit, tstack
+
+__author__ = 'Colour Developers'
+__copyright__ = 'Copyright (C) 2013-2017 - Colour Developers'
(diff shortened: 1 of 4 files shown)


colour/models/rgb/prismatic.py
```diff
@@ -0,0 +1,112 @@
...
+    array([ 0.25...   ,  0.4999999...,  0.75...  ])
+    """
+
+    Lrgb = np.asarray(Lrgb)
+
+    rgb = Lrgb[..., 1:]
+    RGB = (Lrgb[..., 0][..., np.newaxis] /
+           np.max(rgb, axis=-1)[..., np.newaxis] * rgb)
+    # Handling *NaNs* and *infs*.
+    RGB = np.nan_to_num(RGB)
```
MichaelMauderer: Should this happen inside of this function or should the caller be allowed to determine hat to do with these values?

KelSolaar: Good question, I was wondering and decided that the case would occur often enough that I would rather fix it directly. Now that being said it is not something usual in the API so we can not handle them but this means that we have failure each time 0 is given to the functions.

MichaelMauderer: Hm...maybe zero should be handled extra then.
Right now we transform even invalid input so something that's numeric, which might hide bugs that happen in computations before this function is called.

KelSolaar: Yes, any elegant way in mind to do that?

MichaelMauderer: Not sure about elegance but what about just checking whether the `rgb` has any non-zero values and then returning a zero value instead of doing the computation?
```
is_zero = not rgb.any()
if is_zero:
    ....
```

KelSolaar: Something like that seems to work for the first one:

```python
def RGB_to_Prismatic(RGB):
    RGB = np.asarray(RGB)

    L = np.max(RGB, axis=-1)
    s = np.sum(RGB, axis=-1)[..., np.newaxis]
    one_s = 1 / s
    # Handling zero-division *NaNs*.
    one_s[s == 0] = 0
    r, g, b = tsplit(one_s * RGB)

    return tstack((L, r, g, b))
```

KelSolaar:
I have implemented more conservative code that only check if the denominator is equal to zero, all the other stuff should pass unaffected, no easter eggs :)

PR #323, opened 2017-05-16
Add support for JMh input to CIECAM02 reverse mode.

crowsonkb:
This allows colors in the Luo et al. (2006) colorspaces to be converted back to XYZ. I verified that this works now by going from sRGB to CAM02-UCS and back in my own testing.

### DIFF
--- a/colour/appearance/ciecam02.py
+++ b/colour/appearance/ciecam02.py
@@ -285,13 +285,14 @@ s=2.3603053..., Q=195.3713259..., M=0.1088421..., H=278.0607358..., HC=None)
 
 
 def CIECAM02_to_XYZ(J,
-                    C,
+                    rho,
                     h,
                     XYZ_w,
                     L_A,
                     Y_b,
                     surround=CIECAM02_VIEWING_CONDITIONS['Average'],
-                    discount_illuminant=False):
+                    discount_illuminant=False,
+                    input_correlates='JCh'):
     """
     Converts *CIECAM02* specification to *CIE XYZ* tristimulus values.
 
@@ -301,8 +302,8 @@ def CIECAM02_to_XYZ(J,
     ----------
     J : numeric or array_like
         Correlate of *Lightness* :math:`J`.
-    C : numeric or array_like
-        Correlate of *chroma* :math:`C`.
+    rho : numeric or array_like
+        Correlate of *chroma* :math:`C` or *colourfulness* :math:`M`.
     h : numeric or array_like
         *Hue* angle :math:`h` in degrees.
     XYZ_w : array_like
@@ -316,6 +317,8 @@ def CIECAM02_to_XYZ(J,
         Surround viewing conditions.
     discount_illuminant : bool, optional
         Discount the illuminant.
+    input_correlates : str, optional
+        Specifies the triple of correlates to interpret the input as.
 
     Returns
     -------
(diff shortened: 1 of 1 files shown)


KelSolaar:
@crowsonkb : Yet another thank you :) Great! Only thing that bother me is the variable name `rho`, I'm wondering if there is a better choice for that. What do you think @MichaelMauderer 

Leads me to something we have discussed: I think we should probably get the whole `J, C, h` signature part in a single triplet `JCh` that we would then `tsplit` in its individual components.

crowsonkb:
I chose `rho` because the thing the second (saturation-like) member of the triplet has in common with the other possible second members (for instance `s`) is that they are the radius in a cylindrical coordinate system. I could see, similarly, renaming `J` to `z` if support for converting from `Q` is implemented, as well as `h` to `phi` if support for `H` is implemented.

But all this is at least partially moot if the proposed single-triplet signature is used. You would then have to always `tstack()` the J, C/M, and h that come from the forward transform. But you have to do that anyway to use CAM02-UCS. It would probably be slightly more convenient in the reverse mode; I [had to use `*tstack(JMh)`](https://github.com/crowsonkb/iterm_schemer/blob/master/iterm_schemer/cam02ucs.py#L68) in a call to `CIECAM02_to_XYZ`. No idea what the variable name for this single triplet would be. `triplet`?

KelSolaar:
An alternative that would also make sense would be to ingest a `colour.CIECAM02_Specification` class instance, it would make the roundtrip more symmetrical.

KelSolaar:
@crowsonkb, @MichaelMauderer: I did updated the signature of `colour.CIECAM02_to_XYZ` definition to accept a `colour.CIECAM02_Specification` class instance in #330. It allows to do clean round trips like this:

```python
import numpy as np

import colour

XYZ = np.tile(np.array([19.01, 20.00, 21.78]), (100, 1))
XYZ_w = np.array([95.05, 100.00, 108.88])
L_A = 318.31
Y_b = 20.0
surround = colour.CIECAM02_VIEWING_CONDITIONS['Average']
colour.CIECAM02_to_XYZ(
    colour.XYZ_to_CIECAM02(XYZ, XYZ_w, L_A, Y_b, surround), 
    XYZ_w, L_A, Y_b, surround)
```

Let me know what do you think, if you are happy we can merge #330 and park this PR.

MichaelMauderer:
That looks pretty good. I think the tricky thing, though, is making that in this format there are a lot of parameters that are being ignored.
The round trip property is nice of course.

KelSolaar:
I think that there are more combinations of correlates that we could use to get meaningful input but yeah I agree it seems a bit wasteful. That being said I find it a bit more elegant than the current PR with the correlate that change its type given another argument. If we wanted to support another correlate derivation, we would have to tweak the signature further.

MichaelMauderer:
This might be mostly an issue of documentation: does ist actually matter which data structure we use? We use duck typing, all we need is that the data we get has attributes J, C, h. Whether that's a  `colour.CIECAM02_Specification` or something else doesn't really matter, does it?

Otherwise let's just go for it, having it on one data structure makes the most sense in terms of adaptability.

KelSolaar:
@crowsonkb:  #330 has been merged, I'm closing this one as it covers it. Again thanks a lot for all your contributions.

PR #327, opened 2017-05-17
Fix divide by zero warning in CIECAM02 reverse mode

crowsonkb:
When t=0 in CIECAM02 reverse mode (i.e. the user has input pure black) there will be a division by zero when computing P_1. Since P_1 may not be a scalar, I disable the division warning for that line and manually test for finite elements of P_1. According to "CIECAM02 and Its Recent Developments" (http://4843ec7c-89cf-4d26-a36a-0e40ebc9a3a7.s3.amazonaws.com/9781441961891-c1.pdf) p. 53, if t=0, then a=b=0 and as such I only write to elements of a and b where P_1 is finite.

### DIFF
--- a/colour/appearance/ciecam02.py
+++ b/colour/appearance/ciecam02.py
@@ -813,20 +813,22 @@ def opponent_colour_dimensions_reverse(P_n, h):
     a = np.zeros(hr.shape)
     b = np.zeros(hr.shape)
 
-    b = np.where(np.abs(sin_hr) >= np.abs(cos_hr),
+    b = np.where(np.isfinite(P_1) * np.abs(sin_hr) >= np.abs(cos_hr),
                  (n / (P_4 + (2 + P_3) * (220 / 1403) * (cos_hr / sin_hr) -
                        (27 / 1403) + P_3 * (6300 / 1403))),
                  b)
 
-    a = np.where(np.abs(sin_hr) >= np.abs(cos_hr), b * (cos_hr / sin_hr), a)
+    a = np.where(np.isfinite(P_1) * np.abs(sin_hr) >= np.abs(cos_hr),
+                 b * (cos_hr / sin_hr), a)
 
-    a = np.where(np.abs(sin_hr) < np.abs(cos_hr),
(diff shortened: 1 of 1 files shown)


KelSolaar:
@crowsonkb : Thank you, I will look into that one later this weekend or next week, just coming back from a trip and we are in crunch at work so I'm super busy!

colour/appearance/ciecam02.py
```diff
@@ -1344,7 +1346,8 @@ def P(N_c, N_cb, e_t, t, A, N_bb):
     A = np.asarray(A)
     N_bb = np.asarray(N_bb)
 
-    P_1 = ((50000 / 13) * N_c * N_cb * e_t) / t
+    with np.errstate(divide='ignore'):
```
KelSolaar: Any reason to specifically ignore the division error here?

We tend to not mute any warnings by default however there is a global switch in the API to mute them if desired: [colour.filter_warnings](https://github.com/colour-science/colour/blob/develop/colour/utilities/verbose.py#L144)

and we have a bunch of decorators here too: https://github.com/colour-science/colour/blob/develop/colour/utilities/common.py

Everything else looks good!

crowsonkb: I wanted to keep the number of warnings I suppressed to a minimum so I would be able to avoid adding more manual checks for inf/nan than necessary.

KelSolaar: I think I would prefer not *muting* it here, @MichaelMauderer what do you think?

MichaelMauderer: I think warnings should only be muted if their effects are explicitly handled and invisible to the caller.

KelSolaar: Yes agreed.

I haven't *played* the whole stack of @crowsonkb's PRs together but shouldn't #328 ensure that `t` never gets down to 0, and thus making that ignore statement not even needed.

KelSolaar:
Merging that one to move forward, I'll remove the ignore statement in a later commit.

PR #328, opened 2017-05-17
Fix divide by zero in ciecam02.py temporary_magnitude_quantity_reverse()

crowsonkb:
Ensure J cannot be below 1e-16 in temporary_magnitude_quantity_reverse().

### DIFF
--- a/colour/appearance/ciecam02.py
+++ b/colour/appearance/ciecam02.py
@@ -1177,7 +1177,7 @@ def temporary_magnitude_quantity_reverse(C, J, n):
    """
 
     C = np.asarray(C)
-    J = np.asarray(J)
+    J = np.maximum(np.asarray(J), np.finfo(np.float_).eps)
     n = np.asarray(n)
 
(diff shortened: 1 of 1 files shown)


KelSolaar:
@crowsonkb : Thanks!

What about using `np.finfo(np.float_).resolution` in place of *1e-16*. On my platform it is actually *1.0000000000000001e-15*

@MichaelMauderer : What do you think?

crowsonkb:
That seems reasonable. I was trying for something easy to type that was somewhere around float64 machine epsilon (~2.220e-16). That and 1e-8 is my usual epsilon when working with float32 data and 16 is twice that.

MichaelMauderer:
Just had a look what might be the elegant solution and came across this: https://stackoverflow.com/questions/38477908/smallest-positive-float64-number

KelSolaar:
`np.finfo(np.float_).eps` looks good!

KelSolaar:
That could be a good opportunity to amend that one too: https://github.com/colour-science/colour/blob/develop/colour/constants/common.py#L38

KelSolaar:
Thanks a lot!

PR #329, opened 2017-05-18
Correct Luo et al. (2006) publication year in docstrings

crowsonkb:
It was erroneously specified as 2016 thereout.


KelSolaar:
Thank you @crowsonkb!

PR #330, opened 2017-05-29
PR: Update "colour.CIECAM02_to_XYZ" definition signature to use "colour.CIECAM02_Specification" class.

KelSolaar:
References #323.

### DIFF
--- a/colour/appearance/ciecam02.py
+++ b/colour/appearance/ciecam02.py
@@ -339,6 +339,12 @@ def CIECAM02_to_XYZ(CIECAM02_specification,
     XYZ : ndarray
         *CIE XYZ* tristimulus values.
 
+    Raises
+    ------
+    ValueError
+        If neither *C* or *M* correlates have been defined in the
+        `CIECAM02_specification` argument.
+
     Warning
     -------
     The output range of that definition is non standard!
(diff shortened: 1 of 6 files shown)


crowsonkb:
So instances of CIECAM02_Specification are now easy for the user to create, and the reverse mode uses JCh if C is present, otherwise JMh? This looks reasonable...

KelSolaar:
Yes, that is pretty much it, we can tune/alter the logical behaviour at the entrance of the function to whatever we feel is the best.

KelSolaar:
I added a new `colour.as_namedtuple` definition that attempts to convert given *array_like*, *Mapping* or *Numpy* structured array to *namedtuple*, thus it should be even easier to feed `colour. CIECAM02_to_XYZ` definition.

MichaelMauderer:
Looks good, but it seems we are not testing lines 375 and 377.

KelSolaar:
Good catch! I wanted to do it, was not sure how to inject that in the fixtures, decided to implement in priority `colour.as_namedtuple` and completely forgot going back to it.

KelSolaar:
So basically we have to find a smart way, i.e. keeping things elegant, to run two tests by switching `C` here for `M`: https://github.com/colour-science/colour/blob/e5648db9148a7f62863e8832e01a43d6a83c6360/colour/appearance/tests/tests_ciecam02.py#L110-L110

Something I was thinking was to maybe have multiple `ColourAppearanceModelTest.output_specification_from_data` methods we could iterate on here: https://github.com/colour-science/colour/blob/cdbffd063b0c44bb32d752b01647137871434851/colour/appearance/tests/common.py#L133-L133

We could find them through pattern matching them while introspecting the object methods. What do you think?

KelSolaar:
Well, turns out that `ColourAppearanceModelTest.output_specification_from_data` method is reimplemented for CIECAM02 reverse model, made things much easier. I will wait for the unit tests result and merge the PR accordingly as you approved the changes.

PR #332, opened 2017-07-03
PR: Format codebase with "Yapf".

KelSolaar:
Overall the output is quite good, as discussed I'm not a huge fan the way iterables are handled at the moment but I'm crossing fingers that it will be tunable in the future.

A neat side effect is that [the scrutinizer-ci score](https://scrutinizer-ci.com/g/colour-science/colour/inspections/702565f5-969b-4729-9c71-60b346eab846) rised from 9.29 to 9.40.


MichaelMauderer:
Does it come with a tool that could run through Travis to check compliance?
Also is there a configuration that could be imported in Pycharm to check/use this formatting?

KelSolaar:
> Does it come with a tool that could run through Travis to check compliance?

Yes and no, I was originally thinking about that and considering it is quite long to reformat the whole codebase and scrapped the idea of having that running on _travis-ci_, now that being said we have a tool that somehow check the formatting is at least partially correct: **flake8**, **Yapf** will honour flake8 expectations, and has not produced so far any non-compliant output (which is not the case of PyCharm code formatter by the way). It means that as of now we will have to reformat the entire codebase on our end from time to time, which I was doing.

> Also is there a configuration that could be imported in Pycharm to check/use this formatting?
Yes there is a plugin for PyCharm:
https://plugins.jetbrains.com/plugin/9705-yapf-pycharm

Some minor caveats though:

- It does not run on an entire directory in PyCharm by the look of it, I have been doing that from the shell as follows (will add on Website): 

> find /Users/kelsolaar/Documents/Development/colour-science/colour-type f -name '*.py' -exec yapf -p -i {} \;

You can however format the current file, e.g. CTRL + ALT + L on Mac Os X and it is available in the _Code_ menu: **Reformat Code (YAPF)**

- It does modify the file externally, i.e. Undoing == PyCharm reloading the file on disk which is surprising first time.

PR #352, opened 2017-09-30
fix links

ofek:


KelSolaar:
Thanks @ofek!

ofek:
My pleasure! Cool project :)

Issue #335, opened 2017-08-09
Investigate alternative continuous function for sparse spectral data representation.


KelSolaar:
As it occurred when starting to implement #192, our spectral data representation is not great even though it works. The current dict based underlying data structure is limiting and we have been investigating ways to represent our sparse spectral data using continuous functions so that it is possible to compute the range value in the supported domain.

The tested implementation is based on an extrapolator function wrapping an interpolating function and seems to work well albeit very much work in progress and without clues about performance: https://github.com/colour-science/colour-ramblings/blob/master/ramblings_modules/spectral_signal.py

This is one of the two major remaining blocks, the other being #267 that we aim to complete before being able to have Colour reach the stable status.

KelSolaar:
The initial implementation is in progress here: https://github.com/colour-science/colour/tree/feature/continuous_signal/colour/continuous

KelSolaar:
And this is a wrap party, thanks @MichaelMauderer for reviewing the PR!

PR #355, opened 2017-11-07
PR: Continuous Signal

KelSolaar:
This PR implements the alternative spectral data representation and directly references #335.

I think/hope it address a lot of issues with the current dictionary based implementation. There are a few points still to talk about, they might be addressed later but still here they are:

1. Some methods might need to use tolerance/rounding to lessen the strictness of some Numpy definitions, e.g. anything involving equality, `np.in1d` and `np.setxor1d`. The numpy/numpy#7784 issue on Numpy's end is worth reading. The area in the new implementation that I was thinking about are as follows:
- https://github.com/colour-science/colour/blob/80d3703fbed2e9854224ae0ae2945d25c91410e6/colour/continuous/signal.py#L754
- https://github.com/colour-science/colour/blob/80d3703fbed2e9854224ae0ae2945d25c91410e6/colour/continuous/signal.py#L707
- https://github.com/colour-science/colour/blob/80d3703fbed2e9854224ae0ae2945d25c91410e6/colour/continuous/signal.py#L992
2. I would be keen on moving the `colour.SpectralPowerDistribution.extrapolate`, `colour.SpectralPowerDistribution.interpolate`, `colour.SpectralPowerDistribution.align` and `colour.SpectralPowerDistribution.trim` methods (and their `colour.MultiSpectralPowerDistribution` siblings) into `colour.Signal` and `colour.MultiSignal` classes, it however implies we need to handle the `colour.SpectralShape` class, maybe move it in the new `colour.continuous` package and call it `colour.SignalShape`?
3. @MichaelMauderer was talking about implementing a metric that is based on a maximum expected error to the actual perfectly reconstructed signal.

Voila!


colour/algebra/interpolation.py
```diff
@@ -31,23 +37,598 @@
...
     'LinearInterpolator', 'SpragueInterpolator', 'CubicSplineInterpolator',
-    'PchipInterpolator', 'lagrange_coefficients'
+    'PchipInterpolator', 'NullInterpolator', 'lagrange_coefficients'
 ]
 
 
+def kernel_nearest_neighbour(x):
+    """
+    Returns the *nearest-neighbour* kernel evaluated at given samples.
+
```
MichaelMauderer: Should we add a more verbose explanation of what this means?

MichaelMauderer: Maybe it might also suffice to have a top-level docstring that explains what the kernels represent and how they are used.

colour/algebra/interpolation.py
```diff
@@ -31,23 +37,598 @@
...
+        """
+
+        if value is not None:
+            assert is_integer(value), '"window" must be an integer!'
+
+            assert value >= 1, '"window" must be equal or superior to 1!'
+
+            self._window = value
+
+            if self._x is not None:
```
MichaelMauderer: If these self-assignments happen for their side effects, should they be wrapped in a function for documentation purposes?

colour/algebra/interpolation.py
```diff
@@ -31,23 +37,598 @@
...
+    def kernel(self, value):
+        """
+        Setter for the **self.kernel** property.
+        """
+
+        if value is not None:
+            assert hasattr(
+                value,
+                '__call__'), ('"{0}" attribute: "{1}" is not callable!'.format(
+                    'kernel', value))
```
MichaelMauderer: Should we have an `is_kernel` function that hides this check (and maybe tests additional functionality the kernel should provide)?

colour/algebra/interpolation.py
```diff
@@ -31,23 +37,598 @@
...
+    @padding_args.setter
+    def padding_args(self, value):
+        """
+        Setter for the **self.padding_args** property.
+        """
+
+        if value is not None:
+            assert isinstance(value, (dict, OrderedDict)), (
+                '"{0}" attribute: "{1}" type is not "dict" or "OrderedDict"!'
+            ).format('padding_args', value)
```
MichaelMauderer: Does it have to be a `dict` or `OrderedDict` wouldn't any mapping type do?

colour/algebra/interpolation.py
```diff
@@ -511,42 +1100,358 @@ class PchipInterpolator(scipy.interpolate.PchipInterpolator):
...
+        Independent :math:`x` variable values corresponding with :math:`y`
+        variable.
+    y : ndarray
+        Dependent and already known :math:`y` variable values to
+        interpolate.
+    absolute_tolerance : numeric, optional
+        Absolute tolerance.
+    relative_tolerance : numeric, optional
+        Relative tolerance.
+    default : numeric, optional
```
MichaelMauderer: If it's optional, what is the default `default`? (This might be important for all other optional parameters)

MichaelMauderer: Or is this information automatically included in the docs from the signature?

colour/algebra/interpolation.py
KelSolaar: Shall we put that documentation in the `colour.KernelInterpolator` class?

KelSolaar: Maybe based on that: 
image

colour/algebra/interpolation.py
```diff
@@ -31,23 +37,598 @@
...
+    ----------
+    x : array_like
+        Independent :math:`x` variable values corresponding with :math:`y`
+        variable.
+    y : array_like
+        Dependent and already known :math:`y` variable values to
+        interpolate.
+    window : int, optional
+        Width of the window in samples on each side.
+    kernel : callable, optional
```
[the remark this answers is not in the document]
KelSolaar: Well except for the fact the output array shape must match the input one, not really.

colour/algebra/interpolation.py
```diff
@@ -31,23 +37,598 @@
...
+
+            self._y = value
+
+            if self._window is not None:
+                self._y_p = np.pad(self._y, **self._padding_args)
+
+    @property
+    def window(self):
+        """
+        Getter and setter property for the half window.
```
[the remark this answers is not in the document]
KelSolaar: Good catch! It is a a left-over of when that attribute was called `half_window`.

colour/algebra/interpolation.py
KelSolaar: It is not exactly a self-assignement and this is part of the protected API people should not touch. We could maybe implement a `KernelInterpolator._pad` protected method to make more obvious what is happening?

colour/algebra/interpolation.py
KelSolaar: Good point, I think checking for `collections.Mapping` should be enough, I'll update the check.

colour/algebra/interpolation.py
KelSolaar: The built documentation includes all the default values in the definition signature, e.g.: https://colour.readthedocs.io/en/latest/colour.algebra.interpolation.html#colour.algebra.interpolation.lagrange_coefficients

I find it to be a bit redundant to put defaults there (and they must be maintained too).

KelSolaar:
Thanks for starting the review @MichaelMauderer!

colour/algebra/interpolation.py
KelSolaar: E.g.:

```python
class KernelInterpolator(object):
    """
    Kernel based interpolation of a 1-D function.

    The reconstruction of a continuous signal can be described as a linear
    convolution operation. Interpolation can be expressed as a convolution of
    the given discrete function :math:`g(x)` with some continuous interpolation
    kernel :math:`k(w)`: [1]_ [2]_

    :math:`\hat{g}(w_0) = [k * g](w_0) = \sum_{x=-\infty}^{\infty}k(w_0 - x)\cdot g(x)`

    [...]
    """
```

KelSolaar:
@MichaelMauderer: I annoyingly fixed the CI issues related to flake8 and the `I` vs `l` characters in one of the commit of this PR. I think I will have to extract those changes and put them back in *develop* so that all the other branches can benefit from it. This means I will have to re-write history in this branch, I'll wait for your comments regarding my above answers and act accordingly.

colour/algebra/interpolation.py
MichaelMauderer: Ahh, now I get it. It's *that* kind of kernel. That took longer than it probably should have. Yeah, in that case, I think pointing to the mathematical nature of the kernel functions and their requirements (being functions of the type float->float with Domain -inf-->+inf but needs to support arrays, will be used in convolution operation) should be quite helpful.  I like your doc sample.

colour/algebra/interpolation.py
MichaelMauderer: Okay. Do you think an in-line comment then for future internal changes might be helpful anyway?

colour/algebra/interpolation.py
MichaelMauderer: Okay, yeah, I agree this should be fine with the updated docs

colour/algebra/interpolation.py
MichaelMauderer: Yes, agree in that case there's no benefit to that.

colour/algebra/interpolation.py
KelSolaar: Technically the method could be extended to n-dimensional arrays, e.g. images, but we don't really need it now.

MichaelMauderer:
@KelSolaar I think I've addressed all the new comments. Feel free to rewrite history and then I'll continue the review.

colour/algebra/interpolation.py
KelSolaar: Yes, I will add one.

KelSolaar:
Thanks! I'll push the history modified branch tonight.

KelSolaar:
It was simple enough, so I pushed it :)

colour/characterisation/tests/test_cameras.py
```diff
@@ -26,31 +24,7 @@ class TestRGB_SpectralSensitivities(unittest.TestCase):
...
-
-        required_methods = ('__hash__', '__getitem__', '__setitem__',
-                            '__iter__', '__contains__', '__len__', '__eq__',
-                            '__ne__', '__add__', '__sub__', '__mul__',
-                            '__div__', 'get', 'extrapolate', 'interpolate',
-                            'align', 'zeros', 'normalise', 'clone')
-
-        for method in required_methods:
-            self.assertIn(method, dir(RGB_SpectralSensitivities))
+    pass
```
MichaelMauderer: Is there a specific reason for this empty test class?

colour/colorimetry/spectrum.py
```diff
@@ -554,3268 +454,1376 @@ def range(self):
...
+         [ 600.            0.136    ...]]
+        """
+
+        self_shape = self.shape
+        s_e_i = zip((shape.start, shape.end, shape.interval),
+                    (self_shape.start, self_shape.end, self_shape.interval))
+        shape = SpectralShape(
+            * [x[0] if x[0] is not None else x[1] for x in s_e_i])
+        # Defining proper interpolation bounds.
+        # TODO: Provide support for fractional interval like 0.1, etc...
```
MichaelMauderer: Is it worth adding a warning if a fractional bound is encountered here?

colour/utilities/array.py
```diff
@@ -677,3 +732,82 @@ def linear_conversion(a, old_range, new_range):
     out_min, out_max = tsplit(new_range)
 
     return ((a - in_min) / (in_max - in_min)) * (out_max - out_min) + out_min
+
+
+def fill_nan(a, method='Interpolation', default=0):
+    """
+    Fills given array NaNs accordingly to given method.
```
MichaelMauderer: Should this be `according to`? I think I saw this some other places, too.

colour/characterisation/tests/test_cameras.py
KelSolaar: I was holding those tests in case I would need to update them further down the line, I think we should remove them now.

colour/utilities/array.py
KelSolaar: I will do that post-merge on the *v0.3.11* branch, adding a note in Asana about it.

KelSolaar:
Latest code review changes are implemented!

MichaelMauderer:
Should be ready to (squash) and go then!

MichaelMauderer:
Merged!

KelSolaar:
Woot! 2.5 years almost in the making :)

PR #360, opened 2017-11-23
PR: Add references to "BT.470", "BT.601", "SMPTE 240M", review their related implementations and RGB colourspace names.

### LINKED ISSUE #341: Add references to "BT.470", "BT.601", "SMPTE 240M", review their related implementations and RGB colourspace names.
I would be keen on adding some more references regarding *BT.470* and *BT.601*. I think we have most of the stuff covered but I would be keen to improve consistency, add more implementations and/or aliases:

- [x] Implement *BT.601* *OETF*, same than *BT.709* but might require a separate implementation, this is maybe the occasion to uniformise the code with *BT.2020*.
- [x] The Pal/Secam RGB, NTSC RGB and SMPTE-C RGB colourspaces cover the primaries but we might want clearer names tailored to the various flavours, e.g. mention of the standard in the name of new colourspaces and/or line numbers: 525 vs 625. There is a also a very annoying M/Pal flavour that has the same primaries than M/NTSC but not the other Pal versions.
- [x] Implement support for SMPTE 240M colourspace.
- [x] Review RGB colourspace names, e.g. SMPTE-C RGB.
- ~~[ ] Implement aliases for *BT.2100* transfer functions.~~

I think doing so would be helpful from a consistency and documentation perspective. @KevinJW, @Nick-Shaw, I know this is looking back in the past but what is your take on that one?

By the I don't mind much code duplication if it follows the standard.

References
---
- https://www.itu.int/dms_pubrec/itu-r/rec/bt/R-REC-BT.601-7-201103-I!!PDF-E.pdf
- http://www.itu.int/dms_pubrec/itu-r/rec/bt/R-REC-BT.470-6-199811-S!!PDF-E.pdf
- https://www.itu.int/dms_pubrec/itu-r/rec/bt/R-REC-BT.2100-1-201706-I!!PDF-E.pdf

KelSolaar:
Closes #341.


KelSolaar:
I will merge this one as it is essentially name changes and I have a lot of conflicts already to handle because of advances in *develop* branch, #361 is more sensitive though.

PR #368, opened 2017-12-19
PR: Improve "Canon Log" and "Panasonic V-Log" curves implementation and consistency.

### LINKED ISSUE #343: Improve "Canon Log" and "Panasonic V-Log" implementations and consistency
Following on from #336 and the changes implemented in PR #338 for "Sony S-Log", it would improve consistency if the same things were done for "Canon Log" and "Panasonic V-Log".

Like the Sony documentation, both Canon and Panasonic show tables in their documents which give equivalents of 0%, 18% and 90% reflectance as both IRE(%) and 10-bit code values. With ALEXALogC, for example, where 18% reflectance is coded as 10-bit code value 400 (in a 10-bit decode of an ARRIRAW image of a grey card) the IRE(%) on the SDI output of the camera is simply `100 * 400/1023 = 39.1`. For V-Log, the Panasonic table shows 128 as the 10-bit code value for black, and 7.3% as the IRE. To get 7.3% you need to do `100 * legal_to_full(128/1023)`.

Therefore I suggest that both Canon and Panasonic log encoding functions should include `out_legal=True` kwargs and `in_legal=True` for the decoding functions. Defaulting to `True` for these would give a result matching the manufacturer supplied IDTs.

While there could be an argument for including full/legal switches for all the transfer functions, to allow for results matching the code values on the SDI outputs of the cameras, in reality I do not believe that either RED or ARRI log material is ever stored or processed in this form. RED and ARRI SDKs produce raw decodes matching full range, and (almost) all ProRes decoders incorporate a legal_to_full scale, hidden from the user, so raw and ProRes recordings match.

For Canon there is an additional issue. Like Sony, Canon refer to 18% reflectance as 20% 'linear IRE'. Therefore, for example, the `CanonLog3_to_linear` function in the IDT requires its output to be multiplied by 0.9 to return 0.18 instead of 0.2 for the Canon Log 3 coding of mid grey. I suggest that the same `out_reflection=True` kwarg as has been added to the Sony functions should also be added for Canon. Currently to match the ACES IDT linearisation, the following is required: `0.9 * log_decoding_CanonLog3( ( y * 1023 - 64 ) / 876 )`, since the current `log_decoding_CanonLog3` function implements only the `CanonLog3 IRE to Linear` subfunction from the IDT.

The fact that the output of the Canon Log encoding functions is 'IRE' (in the common video levels sense, rather than the obtuse 'linear IRE' sense, used confusingly in the same Canon and Sony documents) and should be converted to code value using a full to legal scale is noted in the current docstring, but the 0.9 'reflectance factor' is not.

KelSolaar:
Closes #343.

### DIFF
--- a/colour/models/rgb/transfer_functions/canon_log.py
+++ b/colour/models/rgb/transfer_functions/canon_log.py
@@ -39,6 +39,8 @@ Notes
 from __future__ import division, unicode_literals
 
 import numpy as np
+
+from colour.models.rgb.transfer_functions import full_to_legal, legal_to_full
 from colour.utilities import as_numeric
 
 __author__ = 'Colour Developers'
@@ -55,7 +57,7 @@ __all__ = [
 ]
 
 
-def log_encoding_CanonLog(x):
+def log_encoding_CanonLog(x, bit_depth=10, out_legal=True, in_reflection=True):
     """
     Defines the *Canon Log* log encoding curve / opto-electronic transfer
     function.
@@ -64,32 +66,43 @@ def log_encoding_CanonLog(x):
     ----------
     x : numeric or array_like
         Linear data :math:`x`.
+    bit_depth : int, optional
+        Bit depth used for conversion.
+    out_legal : bool, optional
+        Whether the *Canon Log* non-linear *IRE* data is encoded in legal
+        range.
+    in_reflection : bool, optional
+        Whether the light level :math`x` to a camera is reflection.
 
     Returns
     -------
     numeric or ndarray
         *Canon Log* non-linear *IRE* data.
-    Notes
-    -----
-    -   Output *Canon Log* non-linear *IRE* data should be converted to code
-        value *CV* as follows: `CV = IRE * (940 - 64) + 64`.
 
     Examples
     --------
-    >>> log_encoding_CanonLog(0.20) * 100  # doctest: +ELLIPSIS
+    >>> log_encoding_CanonLog(0.18) * 100  # doctest: +ELLIPSIS
     32.7953896...
     """
 
     x = np.asarray(x)
 
+    if in_reflection:
+        x = x / 0.9
+
     clog_ire = np.where(x < log_decoding_CanonLog(0.0730597),
                         -(0.529136 * (np.log10(-x * 10.1596 + 1)) - 0.0730597),
                         0.529136 * np.log10(10.1596 * x + 1) + 0.0730597)
 
+    clog_ire = clog_ire if out_legal else legal_to_full(clog_ire, bit_depth)
+
     return as_numeric(clog_ire)
 
 
-def log_decoding_CanonLog(clog_ire):
+def log_decoding_CanonLog(clog_ire,
+                          bit_depth=10,
+                          in_legal=True,
+                          out_reflection=True):
     """
     Defines the *Canon Log* log decoding curve / electro-optical transfer
     function.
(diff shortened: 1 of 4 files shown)


KelSolaar:
@Nick-Shaw : I think the implementation is correct, as consequence of the changes is that the *Canon Log* curves will have their output changed.

Nick-Shaw:
And also references to the input and output values of the Canon functions being `non-linear *IRE* data` should be removed, as this may not be the case. 'IRE' means full range, which we now have switches for.

KelSolaar:
> In fact I think that the switches for the Canon functions are working the wrong way round.

Looking back at the CTL code I think so too. Taking *Canon_EOS_C300_IDT_Ver.1.0/Canon_EOS_C300_IDT_A_Tng_Ver.1.0.ctl*, it ingests legal range CLog data and immediately converts it to full range:

```
	// CodeValue to IRE, ire = (cv-64)/(940-64)
	float iR;
	float iG;
	float iB;
	iR = (rIn*1023 - 64) / 876;
	iG = (gIn*1023 - 64) / 876;
	iB = (bIn*1023 - 64) / 876;
```

we are doing the reverse in `log_decoding_CanonLog`:

```python
    clog = clog if in_legal else full_to_legal(clog, bit_depth)
```

KelSolaar:
So unit tests are breaking, took me a little while to figure out but the root cause is here: https://github.com/colour-science/colour/blob/e6a9bfa93fc8bbd34bf150ca6ef333ed6e606cb8/colour/models/rgb/transfer_functions/canon_log.py#L272.

Maybe I'm overlooking something but this is actually valid for any of the definitions where we have put the new logic: we don't give any parameters to the comparison calls and I think there is/might be misalignment with the input data encoding. As a result we might probably need to put arguments just for those comparisons or have expectations on how the data should be encoded.

As an example, in order to perform a round-trip:

```
a = np.linspace(0, 1, 10000000)
np.testing.assert_almost_equal(
    log_decoding_CanonLog3(log_encoding_CanonLog3(a)), a)
```

I need to modify `log_encoding_CanonLog3` as follows:

```
    clog3 = np.select(
        (x < log_decoding_CanonLog3(0.04076162, bit_depth, False, False),
         x <= log_decoding_CanonLog3(0.105357102, bit_depth, False, False),
         x > log_decoding_CanonLog3(0.105357102, bit_depth, False, False)),
```

Interestingly `log_encoding_CanonLog` and `log_encoding_CanonLog2` seem to be not affected. I'll need to continue investigating.

Nick-Shaw:
`in_legal` and `in_reflectance` both needing to be `False` makes sense, as that makes it work like the original sub-function from the IDT. So that's what's needed to match the thresholds for the inverse. But I would expect that to apply for all three.

KelSolaar:
>  as that makes it work like the original sub-function from the IDT.

Yes, this is what I found, I have updated the definitions accordingly.

Nick-Shaw:
For CanonLog and CanonLog2 the threshold is at linear zero for mirroring, so `out_reflectance` becomes irrelevant. But I would still expect `in_legal` to need to be `False` for it to hit zero.

Nick-Shaw:
Ah. Your round trip test is working on 0-1 linear, so not testing negatives ( or >1). If you reverse it, you would be checking that it round trips the whole log 0-1 range. That would be better.

Nick-Shaw:
Seems like the lowest break point in Canon Log 3 is not happening in exactly the same place for the encode and decode functions. I am investigating.

Nick-Shaw:
I think there is a typo in the lowest segment (both ways). My suggested functions are:
```    clog3 = np.select(
        (x < -0.014,
         x <= 0.014,
         x > 0.014),
        (-0.42889912 * np.log10(-x * 14.98325 + 1) + 0.07623209,
         2.3069815 * x + 0.073059361,
         0.42889912 * np.log10(x * 14.98325 + 1) + 0.069886632))```
and
```    x = np.select(
        (clog3 < log_encoding_CanonLog3(-0.014, 10, False, False),
         clog3 <= log_encoding_CanonLog3(0.014, 10, False, False),
         clog3 > log_encoding_CanonLog3(0.014, 10, False, False)),
        (-(10 ** ((0.07623209 - clog3) / 0.42889912) - 1) / 14.98325,
         (clog3 - 0.073059361) / 2.3069815,
         (10 ** ((clog3 - 0.069886632) / 0.42889912) - 1) / 14.98325))```
I have also switched around the break points to be exactly +/-0.014 in the encoding function.

KelSolaar:
Nicely spotted again @Nick-Shaw! Thanks.

KelSolaar:
As usual, thanks Nick!

PR #372, opened 2017-12-26
PR: Implement support for "CAM16" colour appearance model.

### LINKED ISSUE #334: Implement support for "CAM16" colour appearance model.
References
---

- http://onlinelibrary.wiley.com/doi/10.1002/col.22131/full
- https://www.osapublishing.org/DirectPDFAccess/18E5001A-E9EB-A82F-EFA154AE2E95203D_368272/oe-25-13-15131.pdf?da=1&id=368272&seq=0&mobile=no

KelSolaar:
Closes #334.


colour/appearance/tests/fixtures/cam16.csv
```diff
@@ -0,0 +1,5 @@
+Case,X,Y,Z,X_w,Y_w,Z_w,L_A,Y_b,F,c,N_c,J,C,h,s,Q,M,H
```
MichaelMauderer: Where is the test data from? Should we add a source?

KelSolaar: Oh this is generated as of now, which is very unfortunate. I sent a mail to M. R. Luo start of January but did not hear back from him. I will maybe check if Stephen Westland can get a hook on him.

MichaelMauderer: That's okay I guess, but we should probably document that somewhere.

KelSolaar: Not that I'm aware of unfortunately. Any particular place you would see a mention about that?

KelSolaar:
@MichaelMauderer : I added a `TODO` in the unit tests module: 545dfe9

KelSolaar:
Thanks @MichaelMauderer!

Issue #359, opened 2017-11-22
About Ocean Optics, Thorlabs, Edmund spectrophotometer IO for measurements?


brandondube:
colour, to the best of my knowledge, does not provide the ability to e.g. read a .txt file from an ocean optics or thorlabs or edmund (etc) spectrometer and look at the data as spd, on CIELUV, etc.

Is this considered outside scope for colour, or would pull request(s) that add this functionality be welcome?

KelSolaar:
Hi @brandondube,

You are most likely right! I have not looked into the file formats those devices use, are there any specifications around? For reference X-Rite uses CGATS and it is planned to be implemented as per #354.

brandondube:
@KelSolaar -- unfortunately, the norm optical instrumentation is that each company and instrument tends to uses its own proprietary format.  Sometimes, the format changes between generations of the product as well, though at least with the ocean spectrometers this hasn't happened in a very long time.

I could provide a parser (one I wrote that has an MIT license is here -- https://github.com/brandondube/prysm/blob/master/prysm/io.py#L7 ) as well as some sample data files (~30kb/ea, .txt) if that would be helpful.

colour seems to use a dictionary format for data -- given one of my package's `Spectrum` objects, it's easy to make a dict of the data:

`spectrum_dict = dict(zip(spectrum.wavelengths, spectrum.values))`

A little bit of work/preprocessing is needed depending on the data type (transmission vs raw DN, etc) but it is relatively minimial.

FWIW, my motivation for this is quantifying the color contribution of lenses in a way that is more robust than the ISO version of CCI, which uses only 3 wavelengths, using CIELUV coordinates.

KelSolaar:
That sounds good with me!

> Sometimes, the format changes between generations of the product as well, though at least with the ocean spectrometers this hasn't happened in a very long time.

Do they have version attribute in the file we could rely on?

> colour seems to use a dictionary format for data

So this is actually changing as per colour-science/colour/pull/355, we are working on a new backend that uses `ndarray` directly for domain / range thus you should be able to initialise the classes directly with those two arrays or even more: https://github.com/colour-science/colour/blob/feature/continuous_signal/colour/continuous/signal.py#L106

> FWIW, my motivation for this is quantifying the color contribution of lenses in a way that is more robust than the ISO version of CCI, which uses only 3 wavelengths, using CIELUV coordinates.

That is very cool and I must admit it is the first time I hear about the Colour Contribution Index. Are you basing your new approach from a publication?

[...]

PR #373, opened 2017-12-28
new ocean optics file reader

brandondube:
See #359 

The written function runs under py3, need travis, appveyor, or a human to see if it runs under py2 :)  The docstring test will need to be completed too -- SPD objects contain too much hidden magic for me to understand, but I believe the one I create is correct.


KelSolaar:
Cheers @brandondube! Will look at that in the coming days.

brandondube:
I do not wish for my code to be included in colour.

Issue #389, opened 2018-02-28
Implement support for "DIN 99" colour difference.


KelSolaar:
References
-----------
- https://www.researchgate.net/publication/262186291_Colour_difference_formula_study
- https://www.techstreet.com/standards/din-6176?product_id=1054040

PR #390, opened 2018-03-01
PR: Implement support for "DIN 99" colour difference.

KelSolaar:
References #389.


KelSolaar:
@MichaelMauderer : Numbers matches table in ASTM D2244-07, only doubt I have is regarding textiles weightings, we might need to source DIN-6176.

KelSolaar:
Thanks @MichaelMauderer!

PR #392, opened 2018-03-11
PR: Implement support for "OSA UCS" colourspace.

KelSolaar:
References #118.

PR #393, opened 2018-03-18
PR: Implement support for warning traceback and domain inspection.

KelSolaar:
So this is an initial stab at checking domain and warning users while having a very simple interface for that:

- It enables traceback on warnings by reimplementing [warnings.showwarning].(https://docs.python.org/2/library/warnings.html#warnings.showwarning)
- The relevant objects now have domain inspection, it can be disabled entirely by easier invoking `colour.utilities.filter_warnings` definition or setting the `colour.utilities.ColourWarning.DOMAIN_INSPECTION` attribute to `False`.
- The inspections are fast, simple but will generate false positive at times, they are merely a hint anyway.
- The behaviour is for example as follows for checking a variable *a* values for domain [0, 100]:
    - If the variable is a numeric, its value is compared against 1, if the value is greater we issue a warning.
    - If the variable is an array, values are compared against 1, if 90% of the values are greater than 1, we issue a warning.
- We can tune the defaults or implement different checks although it needs to be simple I think.

Warning
---
https://github.com/colour-science/colour/pull/392 must be merged before this one!


KelSolaar:
@MichaelMauderer : I'm closing this PR as the inspection part is being turned into domain-range scales support.

Issue #403, opened 2018-04-19
Should all RED colour spaces use D65?


Nick-Shaw:
https://github.com/colour-science/colour/blob/5f13cb1458da2d4840cd27f93575c30efdbb42e8/colour/models/rgb/dataset/red.py#L84

The legacy RED colour spaces use ACES as the illuminant, but I believe this is purely historical, because the primaries were derived from the OCIO ACES matrices. RED Wide Gamut RGB uses a D65 illuminant, because that came from a paper published by RED.

Would it not make sense, for consistency, to use D65 for all RED spaces, back-calculating so that with Bradford CAT the ACES RGB to RGB matrices matched the OCIO ones?

This would mean that when plotting multiple camera spaces, the RED ones would not "stick out" as having a different white point, which is confusing, and potentially misleading.

Nick-Shaw:
For reference, the primaries I calculated in the past for the RED colour spaces with D65 white were:

DragonColor
Red (0.758655892625, 0.330355348581)
Green (0.294923619819, 0.708053242057)
Blue (0.0859616011972, -0.0458794369685)

DragonColor2
Red (0.758656214203, 0.330355835732)
Green (0.294923887741, 0.708053363184)
Blue (0.144168726894, 0.0503573845986)

REDcolor
Red (0.701058563196, 0.330180975914)
Green (0.298811317317, 0.625169245946)
Blue (0.135038675229, 0.0352617765633)

REDcolor2
Red (0.897407221957, 0.330776225941)
Green (0.296022094526, 0.684635550893)
Blue (0.0997995129126, -0.0230005131634)

REDcolor3
Red (0.702598658615, 0.330185588912)
Green (0.295782235746, 0.689748258389)
Blue (0.111090529109, -0.00433232097099)

REDcolor4
Red (0.70259815466, 0.330185096184)
Green (0.295782328056, 0.689748253957)
Blue (0.144459236517, 0.0508377209888)

KelSolaar:
@Nick-Shaw : I don't have a super strong opinion on that and was thinking about maybe changing to D65.

Did we get an official word on the illuminant by Graeme Natress, I remember him talking about whitepapers for their "colour science"?

Nick-Shaw:
Graeme has confirmed to me that he used D65 and Bradford CAT for all RED colour spaces.

KelSolaar:
Right that is great, let's go for the change then.

PR #405, opened 2018-04-25
Update RED colour spaces to use D65 white.

Nick-Shaw:
References #403


KelSolaar:
LGTM, thanks @Nick-Shaw!

PR #409, opened 2018-05-03
PR: Implement support for polynomial colour correction.

KelSolaar:
References #63.

Closing the above after 4 years, finally!

PR #413, opened 2018-05-18
Implement support for LUT.

KelSolaar:


KelSolaar:
@Nick-Shaw : Putting it there for initial public review, it depends on #409 which will have to be merged before.

Nick-Shaw:
What about a `1D_to_2D` method which makes 3 identical columns, and is invoked automatically when trying to write a 1D LUT to a .cube?

KelSolaar:
Merging this one @Nick-Shaw! Let's open an enhancement issue with stuff we want to add, e.g. https://github.com/colour-science/colour/pull/413#issuecomment-390774952.

Issue #433, opened 2018-09-18
Implement support for "Filmic Pro 6" log encoding curve.


KelSolaar:
References
---
- http://www.filmicpro.com/FilmicProUserManualv6.pdf

PR #434, opened 2018-09-18
PR: Implement support for "Filmic Pro 6" log encoding curve.

KelSolaar:
References #433.

This is a very poor curve in its current state but given it seems to be used quite a lot on iOS (and I did use it today at work), better have it there! Their fit precision is poor and it is non-invertible. I'm using a *LUT* for inversion, should be precise enough at 7 digits.

### DIFF
--- a/BIBLIOGRAPHY.bib
+++ b/BIBLIOGRAPHY.bib
@@ -337,6 +337,13 @@ title = {{MunsellAndKubelkaMunkToolboxApr2014 - MunsellSystemRoutines/LinearVsRa
 url = {https://github.com/colour-science/MunsellAndKubelkaMunkToolbox},
 year = {2014}
 }
+@misc{Chamberlain2015,
+author = {Chamberlain, Peter},
+title = {{LUT documentation (to create from another program)}},
+url = {https://forum.blackmagicdesign.com/viewtopic.php?f=21&t=40284#p232952},
+urldate = {2018-08-23},
+year = {2015}
+}
 @article{Cheung2004,
 abstract = {The proliferation of low-cost colour imaging devices in the consumer market has led to a greater need to transfer images from one medium or device to another without loss of colour fidelity. A common solution is to characterise each device in terms of its CIE tristimulus values. In this paper two general techniques, artificial neural networks and polynomial transforms, are compared for their usefulness in characterising colour cameras. The neural and polynomial techniques are shown to give approximately similar performance once the parameters of the models are optimised. Since neural networks can be difficult and time-consuming to train, it is concluded that polynomial transforms offer the better alternative for camera characterisation.},
 author = {Cheung, Vien and Westland, Stephen and Connah, David and Ripamonti, Caterina},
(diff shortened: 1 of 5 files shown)


Nick-Shaw:
```
>>> log_encoding_FilmicPro6(log_decoding_FilmicPro6(0.))
8.032950075737677e-06
```
This seems like a very imprecise round trip.

Because the image data in the file will always be ranged 0-1 (or integer equivalent) would there be a case for having the decoding LUT ranged only across the log equivalent of 0-1?

So in `_LOG_DECODING_FILMICPRO_INTERPOLATOR_CACHE`:

``` t = np.arange(0, 1, 0.0001)```

becomes:

```t = np.linspace(0.0021070880715159932, 0.9999971756642045, 10000)```

**Comment by KelSolaar** [comment]
@Nick-Shaw : At that point I don't have a strong opinion about anything :)

Did you solve for 0 and 1 to get the big numbers? If so we can / should probably do it with Numpy directly.

**Comment by KelSolaar** [comment]
@Nick-Shaw : Quietly looking at that from home, what about something like that instead:

```python
from scipy.optimize import fsolve
t = np.linspace(log_encoding_FilmicPro6(np.finfo(DEFAULT_FLOAT_DTYPE).eps), fsolve(lambda x: colour.models.log_encoding_FilmicPro6(x) - 1, 1), 10000)
```

Low bound is given by machine epsilon and high bound by solve.

Nick-Shaw:
I solved for 0 and 1 by iteration, but a more explicit approach included in the code certainly seems preferable.

But in your example t ranges from the log encoding of the machine epsilon, to the linear value which encodes to 1.0. Surely you don't want a range where one end is in the log domain, and the other in the linear? t is the uniform table corresponding to the non-uniform domain, created by log encoding t, for the 1D LUT used by the log decoding function, is it not? The example above gives a low bound of -3.1495804307300799, and the encoding function cannot handle negative numbers. Should it not be simply `np.finfo(DEFAULT_FLOAT_DTYPE).eps` rather than the log encoding of that?

Nick-Shaw:
Perhaps t should simply range from machine epsilon to 1.0, allowing for a round trip of 1.0, even though that encodes to a value >1.

KelSolaar:
> Should it not be simply np.finfo(DEFAULT_FLOAT_DTYPE).eps rather than the log encoding of that?

It is / was the intent, i.e what is the minimal value we can encode without blowing up! :)

I dumbly copy-pasted my interpreter content into the first argument of `np.linspace` but it should indeed have been `t = np.linspace(np.finfo(DEFAULT_FLOAT_DTYPE).eps, fsolve(lambda x: colour.models.log_encoding_FilmicPro6(x) - 1, 1), 10000)`

For the upper bound I wanted to not overflow the container because of precision issues, but we can probably keep 1, although at that point I'm wondering if we should not just keep `[0, 1]`

Nick-Shaw:
You're probably right. Worrying excessively about precision for a format used for 8-bit iPhone recordings is probably pointless! Let's stick to `[0, 1]`.

KelSolaar:
Thanks @Nick-Shaw! It should hopefully be fixed whenever and if they update their curve! :)

Issue #438, opened 2018-09-27
Implement support for LED generation model of "Spectral design considerations for white LED color rendering".


KelSolaar:
References
---
- Ohno, Y. (2005). Spectral design considerations for white LED color rendering. Optical Engineering, 44(11), 111302. doi:10.1117/1.2130694

I actually have implemented the single LED function here: https://colab.research.google.com/drive/1ipug7T0XEAPEjuezMinSIbnFoq29hHG3#scrollTo=mcs2Z0cC8UE9.

It was done as support for the following thread: https://colab.research.google.com/drive/1ipug7T0XEAPEjuezMinSIbnFoq29hHG3#scrollTo=mcs2Z0cC8UE9

Issue #285, opened 2016-09-04
Update colour quality "CQS" implementation to use "NIST CQS 9.0" method.


KelSolaar:
## References
- https://www.researchgate.net/file.PostFileLoader.html?id=5541c498f15bc7cc2c8b4578&assetKey=AS%3A273582771376136%401442238623549

PR #439, opened 2018-09-27
PR: Implement support for the LED generation model of "Spectral design considerations for white LED color rendering".

KelSolaar:
References #438.

This is quite simple and I verified the numbers against *Yoshi Ohno* spreadsheet. Reminds me that once we have released v0.3.12 I need to take care of #285.

### DIFF
--- a/colour/colorimetry/generation.py
+++ b/colour/colorimetry/generation.py
@@ -3,13 +3,29 @@
 Spectral Generation
 ===================
 
-Defines various classes performing spectral generation:
+Defines various objects performing spectral generation:
+
+-   :func:`colour.constant_spd`
+-   :func:`colour.zeros_spd`
+-   :func:`colour.ones_spd`
+-   :func:`colour.gaussian_spd`
+-   :func:`colour.single_led_spd_Ohno2005`
+-   :func:`colour.multi_led_spd_Ohno2005`
 
 See Also
 --------
 `Spectrum Jupyter Notebook
 <http://nbviewer.jupyter.org/github/colour-science/colour-notebooks/\
 blob/master/notebooks/colorimetry/generation.ipynb>`_
+
+References
+----------
+-   :cite:`Ohno2005` : ﻿Ohno, Y. (2005). Spectral design considerations for
+    white LED color rendering. Optical Engineering, 44(11), 111302.
+    doi:10.1117/1.2130694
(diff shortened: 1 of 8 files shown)


colour/colorimetry/generation.py
```diff
@@ -0,0 +1,349 @@
...
+    >>> spd.shape
+    SpectralShape(360.0, 780.0, 1.0)
+    >>> spd[400]
+    1.0
+    """
+
+    return constant_spd(1, shape)
+
+
+def gaussian_spd(peak_wavelength, fwhm, shape=DEFAULT_SPECTRAL_SHAPE):
```
MichaelMauderer: It its Gaussian, should it be specified by the standard Gaussian parameters (mean and std)? Scipy also already uses that format out the box: https://docs.scipy.org/doc/scipy-0.14.0/reference/generated/scipy.signal.gaussian.html

colour/colorimetry/generation.py
```diff
@@ -0,0 +1,349 @@
...
+<http://nbviewer.jupyter.org/github/colour-science/colour-notebooks/\
+blob/master/notebooks/colorimetry/generation.ipynb>`_
+
+References
+----------
+-   :cite:`Ohno2005` : ﻿Ohno, Y. (2005). Spectral design considerations for
+    white LED color rendering. Optical Engineering, 44(11), 111302.
+    doi:10.1117/1.2130694
+-   :cite:`Ohno2008a` : Ohno, Y., & Davis, W. (2008). NIST CQS simulation 7.4.
+Retrieved from https://drive.google.com/file/d/\
```
MichaelMauderer: This link does not copy paste easily from source.

KelSolaar: Yeah unfortunately we don't (did not) have choice or it will break the .rst/.html output.

colour/colorimetry/generation.py
KelSolaar: I can / will add add a `method` argument with for example:

- `Standard`, the default that uses *mean* and *std* (which is how I usually use it btw, and have it implemented in a few notebook.)
- `Ohno (2005)` that uses the current path.

KelSolaar:
Thanks @MichaelMauderer, I will add the `method` argument as per the above, add relevant tests and merge!

PR #441, opened 2018-10-20
Fix invalid azure config

MichaelMauderer:


KelSolaar:
Thanks @MichaelMauderer, I will try to fix the breaking tests in another branch!

PR #465, opened 2019-04-10
Add Colorio to "See Also" section of README

jamadagni:
The Colorspacious Python library is good but it doesn't include more recent models than CAM02-UCS such as CAM-16, Jzazbz, ICtCp etc. Colorio supports these.

The Colorspacious author clearly states inability to update the library:
https://github.com/njsmith/colorspacious/issues/10
https://github.com/njsmith/colorspacious/issues/13

So definitely Colorio should be mentioned above Colorspacious as it is better than the letter.


KelSolaar:
Thanks @jamadagni!

Issue #450, opened 2018-12-14
Update "eotf_BT2100_HLG" to July 2018 BT.2100-2 specification.


nick-shaw:
The black handling in the HLG EOTF and OOTF has been changed with the July 2018 update to BT.2100 (thanks to Tim Borer for alerting me to this.)

For completeness, I suppose we should add a `use_legacy` kwarg, so the previous version can be used.

KelSolaar:
We should probably look at this one sooner than later. I guess the update is the black lift here:

![image](https://user-images.githubusercontent.com/99779/57601401-85250300-75b0-11e9-8afe-3601d4cf1818.png)

References
---
- https://www.itu.int/dms_pubrec/itu-r/rec/bt/R-REC-BT.2100-2-201807-I!!PDF-E.pdf

KelSolaar:
@nick-shaw : I was looking at that and I just noticed there is a negative clamp that might prevent the full reverse implementation, I'm thinking about dropping it.

PR #478, opened 2019-05-20
PR: Implement support for "ITUR BT.2100-2".

KelSolaar:
References #450.

### DIFF
--- a/BIBLIOGRAPHY.bib
+++ b/BIBLIOGRAPHY.bib
@@ -1,11 +1,3 @@
-@misc{AdobeSystems2013,
-  author       = {{Adobe Systems}},
-  title        = {{Adobe DNG Software Development Kit (SDK) - 1.3.0.0
-    -
-    dng_sdk_1_3/dng_sdk/source/dng_temperature.cpp::dng_temperature::Set_xy_coord}},
-  url          = {https://www.adobe.com/support/downloads/dng/dng_sdk.html},
-  year         = 2013,
-}
 @misc{AdobeSystems2013b,
   author       = {{Adobe Systems}},
   keywords     = {Iridas,look-up table,specification},
@@ -20,6 +12,14 @@
   url          = {http://www.adobe.com/digitalimag/pdfs/AdobeRGB1998.pdf},
   volume       = 2704,
   year         = 2005,
+}
+@misc{AdobeSystems2013,
+  author       = {{Adobe Systems}},
+  title        = {{Adobe DNG Software Development Kit (SDK) - 1.3.0.0
+    -
+    dng_sdk_1_3/dng_sdk/source/dng_temperature.cpp::dng_temperature::Set_xy_coord}},
+  url          = {https://www.adobe.com/support/downloads/dng/dng_sdk.html},
+  year         = 2013,
 }
 @misc{AdobeSystems2013a,
   author       = {{Adobe Systems}},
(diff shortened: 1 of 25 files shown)


KelSolaar:
@nick-shaw : I removed the clamp compared to the reference, I was not comfortable having that middle of the functions.

nick-shaw:
> @nick-shaw : I removed the clamp compared to the reference, I was not comfortable having that middle of the functions.

I guess that makes sense. For a real display there is no meaning to an EOTF producing a negative value, whereas when performing calculations for various purposes there may be a reason to preserve negatives. And the user can always clamp negative values themselves.

Not sure it helps with invertibility though, since **_OETF<sup>-1</sup>(x)_** where **_x < 0_** will be the same as **_OETF<sup>-1</sup>(-x)_**, since **_OETF<sup>-1</sup>(x) = x<sup>2</sup> / 3_** when **_x <= 0.5_**.

Might there be a case for replacing `(E_p / r) ** 2` with e.g. `gamma_function((E_p / r), 2, negative_number_handling='mirror'` in `oetf_reverse_ARIBSTDB67` to allow for invertibility?

KelSolaar:
Yes that makes sense, we should also change the square root in the OETF accordingly: `r * np.sqrt(E)`.

I will do the changes.

nick-shaw:
https://github.com/colour-science/colour/blob/d2e2158d5b1f831cdf6e315f7b77572f6cdb6e5c/colour/models/rgb/transfer_functions/itur_bt_2100.py#L471
Shouldn't this be `return oetf_ARIBSTDB67(12 * 0.5, constants=BT2100_HLG_CONSTANTS)` to make use of the calculated constants? Likewise for line 516?

nick-shaw:
It looks like the HLG OOTF has not been changed for BT.2100-2. As far as I can see, both methods use the same `ootf_HLG_BT2100` which is not correct. When you use a non-zero `L_B` the output for zero input should track `L_B`, which it currently doesn't.

```
>>> eotf_HLG_BT2100(0.0, L_B=0.5, L_W=1000,  method='ITU-R BT.2100-2')
0.99975000000000014
```

KelSolaar:
Mmmh, the black lift should be applied once:

```
>>> colour.models.eotf_HLG_BT2100(0.212132034355964, 0.01)
eotf_HLG_BT2100(0.212132034355964, 0.01)
  eotf_HLG_BT2100_2(0.212132034355964, 0.01, 1000, None)
    black_level_lift_HLG_BT2100(0.01, 1000, None)
      gamma_function_HLG_BT2100(1000)
    ootf_HLG_BT2100(0.016635216736461608, 0.01, 1000, None)
      gamma_function_HLG_BT2100(1000)
7.3421242063785908
```

I will double-check tonight.

KelSolaar:
So I totally missed the fact that the OOTF has been also changed:

BT-2100-1
---

image

BT-2100-2
---
image

KelSolaar:
> https://github.com/colour-science/colour/blob/d2e2158d5b1f831cdf6e315f7b77572f6cdb6e5c/colour/models/rgb/transfer_functions/itur_bt_2100.py#L471
> 
> 
> Shouldn't this be `return oetf_ARIBSTDB67(12 * 0.5, constants=BT2100_HLG_CONSTANTS)` to make use of the calculated constants? Likewise for lines 516, 615 and 677?

We put the following comment for `colour.models.rgb.transfer_functions.itur_bt_2100.BT2100_HLG_CONSTANTS` attribute docstring:

```
*Recommendation ITU-R BT.2100* *Reference HLG* constants expressed in their
analytical form in contrast to the *ARIB STD-B67 (Hybrid Log-Gamma)* numerical
reference.

Those constants are not currently in use and are provided as a reference only.
```

So we elected not to use the analytical form, should we change it?

PR #479, opened 2019-05-20
PR: Implement support for Fuji "FLog" and "FGamut" colourspace.

sobotka:
Adds in Fuji F-Log / F-Gamut, without tests.


KelSolaar:
Thanks @sobotka ! Will start to take a look at that. Seems like the tests are breaking on the reciprocity case.

colour/__init__.py
```diff
@@ -908,6 +908,10 @@ def __getattr__(self, attribute):
             'colour.log_decoding_VLog',
             'colour.models.log_decoding_VLog',
         ],
+        [
+            'colour.log_decoding_FLog',
```
KelSolaar: Those API changes are not required here because the definitions have never been renamed or their access point changed: it is the first time they are featured in Colour.

colour/__init__.py
```diff
@@ -988,6 +992,10 @@ def __getattr__(self, attribute):
             'colour.log_encoding_VLog',
             'colour.models.log_encoding_VLog',
         ],
+        [
+            'colour.log_encoding_FLog',
```
KelSolaar: Same here, not needed.

colour/models/rgb/dataset/f_gamut.py
```diff
@@ -0,0 +1,102 @@
...
+
+See Also
+--------
+`RGB Colourspaces Jupyter Notebook
+<http://nbviewer.jupyter.org/github/colour-science/colour-notebooks/\
+blob/master/notebooks/models/rgb.ipynb>`_
+
+References
+----------
+-   :cite:`FujiVer1` : F-Log Data Sheet Ver.1.0.
```
KelSolaar: I can change it but if you want the citation which depends on the entry I added to Mendeley (and will push to the *BIBLIOGRAPHY.bib* file) should be like that:

> -   :cite:`Fujifilm2016` : Fujifilm. (2016). F-Log Data Sheet Ver.1.0. Retrieved from https://www.fujifilm.com/support/digital_cameras/software/lut/pdf/F-Log_DataSheet_E_Ver.1.0.pdf

colour/models/rgb/transfer_functions/fuji_flog.py
```diff
@@ -0,0 +1,201 @@
+# -*- coding: utf-8 -*-
+"""
+Fuji F-Log Log Encoding
+============================
```
KelSolaar: Might need to shave a few `=`, in that direction it is no big deal but if shorter it slightly borks the documentation :)

colour/models/rgb/transfer_functions/fuji_flog.py
```diff
@@ -0,0 +1,201 @@
...
+
+    +------------+-----------------------+---------------+
+    | **Range**  | **Scale - Reference** | **Scale - 1** |
+    +============+=======================+===============+
+    | ``F_out``  | [0, 1]                | [0, 1]        |
+    +------------+-----------------------+---------------+
+
+    References
+    ----------
+    :cite:`FujiVer1`
```
KelSolaar: The citation keys will also need update.

colour/models/rgb/transfer_functions/fuji_flog.py
```diff
@@ -0,0 +1,201 @@
...
+
+    Examples
+    --------
+    >>> log_encoding_FLog(0.18)  # doctest: +ELLIPSIS
+    0.46333651051465585...
```
KelSolaar: Should trim at 7 decimals, the ellipsis is here to help platform differences on numerical computations.

> 0.4633365...

colour/models/rgb/transfer_functions/fuji_flog.py
```diff
@@ -0,0 +1,201 @@
...
+    +------------+-----------------------+---------------+
+
+    References
+    ----------
+    :cite:`FujiVer1`
+
+    Examples
+    --------
+    >>> log_decoding_FLog(0.46333651051465585)  # doctest: +ELLIPSIS
+    0.18000000000000002...
```
KelSolaar: Likewise 7 decimals :)

colour/models/rgb/transfer_functions/fuji_flog.py
```diff
@@ -0,0 +1,201 @@
...
+*Fuji F-Log* colourspace constants.
+
+FLOG_CONSTANTS : Structure
+"""
+
+
+def log_encoding_FLog(L_in,
+                      bit_depth=10,
+                      out_legal=False,
+                      in_reflection=True,
```
KelSolaar: Do we need support for that? Seems like the reference does not specifically mention it.

KelSolaar:
@sobotka : Thanks again, I added a few minor comments that should be quick to address. I will need to spend some time to understand why the current unit tests are breaking.

Two things:

- Would be great to squash the *Fix typo on F_out* and *Update doctest string* commits into *Add F-Log and F-Gamut* when we are done to keep the history a bit tidy.
- We will need to add a `colour/models/rgb/transfer_functions/tests/test_fuji_flog.py` unit test module.

@nick-shaw : Do you think we should drop the `in_reflection` argument?

sobotka:
Rebased against develop's head and squashed everything into a single merge point.

I believe you can change the branch from [sobotka:fuji](/sobotka/colour/tree/fuji) to [sobotka:fuji-revisions](/sobotka/colour/tree/fuji-revisions)?

KelSolaar:
I will close this one and use #480 instead!

PR #489, opened 2019-07-27
PR: Fix "from collections import Mapping, MutableMapping" with Python 3.8.

vidakDK:
In Python 3.8 from collections import Mutable no longer works, instead
from collections.abc import Mutable must be used

### DIFF
--- a/colour/utilities/data_structures.py
+++ b/colour/utilities/data_structures.py
@@ -27,7 +27,7 @@ structures.py#L37
 
 from __future__ import division, unicode_literals
 
-from collections import Mapping, MutableMapping
+from collections.abc import Mapping, MutableMapping
 
 __author__ = 'Colour Developers'
 __copyright__ = 'Copyright (C) 2013-2019 - Colour Developers'


vidakDK:
Ah this needs to work in Python 2 as well? Not sure if worth to fix if Python 2 will be gone in a couple months :)

KelSolaar:
Hi @vidakDK,

I'm not sure yet how long after Python 2 death we will maintain Colour compatibility with it but it will not go instantly.

What about using something like that instead:

```python
import sys

if sys.version_info[:2] >= (3, 8):
    from collections.abc import Mapping, MutableMapping
else:
    from collections import Mapping, MutableMapping
```

KelSolaar:
Looking at this one and seems like there are a few more imports to correct:

```
/colour-science/colour/colour/algebra
    interpolation.py
        from collections import OrderedDict, Mapping
/colour-science/colour/colour/continuous
    multi_signals.py
        from collections import Iterator, Mapping, OrderedDict, Sequence
    signal.py
        from collections import Iterator, Mapping, OrderedDict, Sequence
/colour-science/colour/colour/plotting
    models.py
        from collections import Mapping
/colour-science/colour/colour/utilities
    array.py
        from collections import Mapping
    data_structures.py
        from collections import Mapping, MutableMapping
```

Are you happy to do that @vidakDK, I can do it otherwise?

KelSolaar:
I was hoping that `six` would help here but seeing that this PR does not move, I don't think it will: https://github.com/benjaminp/six/pull/241

KelSolaar:
Closing this one for Hacktoberfest purposes, thanks @vidakDK!

PR #496, opened 2019-08-29
PR: Automatic Colour Conversion Graph

KelSolaar:
I will refine a few things, i.e. more tests, coverage and guards in case NetworkX is not installed, but overall this is it.

https://github.com/colour-science/colour/raw/feature/graph/docs/_static/Examples_Colour_Automatic_Conversion_Graph.png

```python
>>> sd = colour.COLOURCHECKERS_SDS['ColorChecker N Ohta']['dark skin']
>>> convert(sd, 'Spectral Distribution', 'sRGB', verbose_parameters={'describe': 'short'})
===============================================================================
*                                                                             *
*   [ Conversion Path ]                                                       *
*                                                                             *
*   "sd_to_XYZ" --> "XYZ_to_sRGB"                                             *
*                                                                             *
===============================================================================
array([ 0.45675795,  0.30986982,  0.24861924])
>>> illuminant = colour.ILLUMINANTS_SDS['FL2']
>>> convert(sd, 'Spectral Distribution', 'sRGB', sd_to_XYZ={'illuminant': illuminant})
array([ 0.47924575,  0.31676968,  0.17362725])
```


KelSolaar:
Just noticed that:

```
print(colour.convert('#808080', 'Hexadecimal', 'Scene-Referred RGB', verbose_parameters={'describe': 'short'}))

===============================================================================
*                                                                             *
*   [ Conversion Path ]                                                       *
*                                                                             *
*   "HEX_to_RGB" --> "encoding_cctf" --> "decoding_cctf"                      *
*                                                                             *
===============================================================================
[ 0.50196078  0.50196078  0.50196078]
```

While it should be:

```
===============================================================================
*                                                                             *
*   [ Conversion Path ]                                                       *
*                                                                             *
*   "HEX_to_RGB" --> "decoding_cctf"                                          *
*                                                                             *
===============================================================================
[ 0.2158605  0.2158605  0.2158605]
```

KelSolaar:
@MichaelMauderer : I think I will probably add a dedicated _verbose_ definition instead of intertwining the current verbose code into the `colour.convert` definition.

colour/graph/conversion.py
```diff
@@ -0,0 +1,803 @@
...
+    ndarray or numeric or SpectralDistribution
+        Converted object :math:`a`.
+
+    Notes
+    -----
+    -   Various defaults have been systematically adopted compared to the
+        low-level *Colour* API:
+
+        -   The domain-range scale is **'1'**.
+        -   The default illuminant is *CIE Standard Illuminant D Series* *D65*.
```
MichaelMauderer: Is this also the expected/required input illuminant? If so this could merit additional highlighting.

MichaelMauderer:
> @MichaelMauderer : I think I will probably add a dedicated _verbose_ definition instead of intertwining the current verbose code into the `colour.convert` definition.

Sounds good!

Otherwise this seems good to me. Might need extensive testing to make sure it works as expected in all cases.

colour/graph/conversion.py
KelSolaar: It is not required but more the default wherever the underlying definitions require an illuminant. That said it is effectively a good thing to state that the data passed to `colour.convert` is expected to be under **D65**.

KelSolaar:
Thanks @MichaelMauderer ! So as for the tests I have few in mind I would like to add but we cannot really test all the possible paths, there are so many, if you have an idea, let me know!

MichaelMauderer:
Wouold it be feasable to programatically check for a given starting colour that all round-trips work fine? That could already confirm that at least everything is internally consistent.

KelSolaar:
@MichaelMauderer : I was looking at the roundtrip thing and I don't know if there is a way to make that work without an extensive classification of the various paths because:

- Some of them reduce triplets to doublets, e.g. _CIE XYZ_ to _CIE xy_.
- Some of them reduce triplets to singlet: e.g. _RGB_ to _RGB Luminance_.
- Some might not roundtrip at all because of functions not being bijective, e.g. _CCT_ to _CIE xy_ and back.

I would probably be inclined to go for tests that validate the logic of the graph itself, for example conversion from _Hexadecimal_ to _RGB Luminance_ must go through decoding of the RGB values to linear light values. Maybe worth for @nick-shaw to drop an eye on the graph itself.

KelSolaar:
@MichaelMauderer : Any further thoughts on this one? Nick is at IBC and will probably be busy this week but I will talk to him, he might have some ideas.

MichaelMauderer:
Agree, the non-invertible nature of some of the paths makes that difficult. The fact that there are so many different paths is, what makes the graph necessary after all. 

As you say, there are two layers to it: the actual Conversion Paths, and the execution of the conversion. So it makes sense to have some tests for both. 

It's really good that there are functions that there is a verbose option and public functionality for getting information about the path, in case people do encounter issues.

KelSolaar:
Closed the PR by mistake! Do you think we should merge and release it with the caveat that it is beta?

MichaelMauderer:
I think that's a good idea. Is it possible to do something like a
deprecation warning but for "beta"?

KelSolaar:
We could issue a `ColourWarning` and if people are bored, they just filter it and we indicate how to do it in the warning message itself.

MichaelMauderer:
Sounds good!

KelSolaar:
What about:

```
ColourUsageWarning: The "Automatic Colour Conversion Graph" is a beta feature, be mindful of this when using it. Please report any unexpected behaviour and do not hesitate to ask any questions should they arise.
This warning can be disabled with the "colour.utilities.filter_warnings" definition as follows: "colour.utilities.filter_warnings(colour_usage_warnings=True)"
```

MichaelMauderer:
Looks good. Is the proposed way to silence the warning going silence only
this warning or all colour warnings?

KelSolaar:
Ah good point it would effectively silence all the `ColourUsageWarning` based warnings. Let me check what they are so far.

KelSolaar:
That is a lot of them:

```
Targets
Occurrences of 'usage_warning(' in Directory /Users/kelsolaar/Documents/Development/colour-science/colour
Found Occurrences
Unclassified occurrence
colour
/Users/kelsolaar/Documents/Development/colour-science/colour/colour/adaptation
cie1994.py
        usage_warning(('"Y_o" luminance factor must be in [18, 100] domain, '
/Users/kelsolaar/Documents/Development/colour-science/colour/colour/appearance
hunt.py
        usage_warning('Unspecified proximal field "XYZ_p" argument, using '
        usage_warning('Unspecified "N_cb" argument, using approximation: '
        usage_warning('Unspecified "N_bb" argument, using approximation: '
        usage_warning(
        usage_warning(
        usage_warning(
/Users/kelsolaar/Documents/Development/colour-science/colour/colour/blindness
machado2009.py
        usage_warning(
        usage_warning(
/Users/kelsolaar/Documents/Development/colour-science/colour/colour/colorimetry
lightness.py
        usage_warning('"W*" Lightness computation is only applicable for '
/Users/kelsolaar/Documents/Development/colour-science/colour/colour/examples
examples_colour.py
usage_warning('This is a first usage warning and it has been filtered!')
/Users/kelsolaar/Documents/Development/colour-science/colour/colour/graph
conversion.py
    usage_warning(
/Users/kelsolaar/Documents/Development/colour-science/colour/colour/io
image.py
            usage_warning(
            usage_warning(
/Users/kelsolaar/Documents/Development/colour-science/colour/colour/io/luts
iridas_cube.py
        usage_warning('"LUT" is a "LUTSequence" instance was passed, '
lut.py
            usage_warning(
sony_spi1d.py
        usage_warning('"LUT" is a "LUTSequence" instance was passed, '
sony_spi3d.py
        usage_warning('"LUT" is a "LUTSequence" instance was passed, '
/Users/kelsolaar/Documents/Development/colour-science/colour/colour/models/rgb/transfer_functions
__init__.py
        usage_warning(
        usage_warning(
itur_bt_2100.py
        usage_warning(
        usage_warning(
        usage_warning(
        usage_warning(
/Users/kelsolaar/Documents/Development/colour-science/colour/colour/notation
munsell.py
        usage_warning('"{0}" is not within "MacAdam" limits for illuminant '
/Users/kelsolaar/Documents/Development/colour-science/colour/colour/temperature
cie_d.py
        usage_warning(('Correlated colour temperature must be in domain '
hernandez1999.py
    usage_warning('"Hernandez-Andres et al. (1999)" method for computing '
kang2002.py
        usage_warning(('Correlated colour temperature must be in domain '
mccamy1992.py
    usage_warning('"*McCamy (1992)" method for computing "CIE xy" '
/Users/kelsolaar/Documents/Development/colour-science/colour/colour/utilities
deprecation.py
                usage_warning(str(change))
verbose.py
def usage_warning(*args, **kwargs):
Usage in comments
colour
/Users/kelsolaar/Documents/Development/colour-science/colour/colour/utilities
verbose.py
    >>> usage_warning('This is a runtime warning!')  # doctest: +SKIP
    >>> usage_warning('This is an usage warning!')  # doctest: +SKIP
```

KelSolaar:
But we have a context manager, obviously, because were are nice ;) So it would actually be:

```python
with colour.utilities.suppress_warnings(colour_usage_warnings=False):
   convert(...
```

MichaelMauderer:
That should silence only this warning and solve the problem of missing out on all the other ones. Great!

KelSolaar:
It is in the box! :)

KelSolaar:
@MichaelMauderer : Should we merge?

MichaelMauderer:
Let's do it!

PR #503, opened 2019-09-22
PR: Implement support for "Github Actions".

KelSolaar:
References #501.


KelSolaar:
Easiest to see what is going on is to check https://github.com/KelSolaar/colour/actions.

KelSolaar:
Thanks!

That is an excellent point, I thought about it but flagged it as not important, now and retrospectively, I realise that it is probably critical as it would allow us to release more often. I will look at https://www.colour-science.org/contributing/#releasing-colour and see how we can simplify the process.

Merging!

Issue #508, opened 2019-10-01
Deprecate "colour.models.XYZ_to_colourspace_model" definition.


KelSolaar:
With the new Automatic Colour Conversion Graph implemented, we should discourage usage of the `colour.models.XYZ_to_colourspace_model` definition. It has been a convenient wrapper but it should disappear because its functionality is superseded by the `colour.convert` definition.

Deprecating API features is done by using the various objects in the `colour.utilities.deprecation` module. In this case, the [API_CHANGES](https://github.com/KelSolaar/colour/blob/51f69faca23a7ccc20c9bfdca2e3359b8496a044/colour/models/__init__.py#L92) dictionary should be updated with a new key along those lines:

```python
API_CHANGES = {
    'FutureRemove': [
        '`colour.models.XYZ_to_colourspace_model',
    ],
    [...]
}
```

KelSolaar:
@feralpoosum : Thanks :)

Do you mind giving me your first and last name for attribution purposed, e.g. https://github.com/colour-science/colour/blob/develop/CONTRIBUTORS.rst, the Website and Zenodo?

KelSolaar:
I can use _Matt Pppp_ if you are fine with it btw!

PR #509, opened 2019-10-01
Deprecated 'colour.models.XYZ_to_colourspace_model'

feralpoosum:
Not sure if this is all that is needed but deprecated as according to issue #508

### DIFF
--- a/colour/models/__init__.py
+++ b/colour/models/__init__.py
@@ -90,6 +90,9 @@ class models(ModuleAPI):
 
 # v0.3.14
 API_CHANGES = {
+    'FutureRemove': [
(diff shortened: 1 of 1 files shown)


KelSolaar:
Hi @feralpoosum,

Thanks, and yes that should be it! Do you mind just amending the commit to fix the typo please:

`Depcreated` --> `Deprecate`

Cheers,

Thomas

feralpoosum:
Hi @KelSolaar 

I have just realised I put a typo in the branch name as well. I'm not really sure how to change the commit. Would it just be easier to close the PR and I'll open another one

Thanks

KelSolaar:
Sounds good with me, let's do that then!

Issue #505, opened 2019-09-30
Fix missing spaces in "colour.encoding_cctf" and "colour.decoding_cctf" definitions docstrings.


sobotka:
Missing a space after the `the`.

https://github.com/colour-science/colour/blob/2bf80feb352ce88292e904174411023a48b67fa5/colour/models/rgb/transfer_functions/__init__.py#L782

KelSolaar:
Good catch! Actually there is another one two lines above, there must be some other similar in the file.

KelSolaar:
@sobotka : Mind opening a PR for this one please whenever you have some cycles, should be very easy :)

sobotka:
Will do! If you catch others, drop them here and I’ll do one squished commit.

KelSolaar:
Thanks @sobotka! I did not see any but I must admit I'm not actively chasing them, just taming them as I go usually.

PR #514, opened 2019-10-01
PR: Fix missing spaces in various docstrings.

pavithraes:
Fixes issue: #505

### DIFF
--- a/colour/models/rgb/transfer_functions/__init__.py
+++ b/colour/models/rgb/transfer_functions/__init__.py
@@ -777,9 +777,9 @@ def encoding_cctf(value, function='sRGB', **kwargs):
 
     if 'itu-r bt.2100' in function.lower():
         usage_warning(
-            'With the "ITU-R BT.2100" method, only the inverse electro-optical'
+            'With the "ITU-R BT.2100" method, only the inverse electro-optical '
             'transfer functions (EOTFs / EOCFs) are exposed by this '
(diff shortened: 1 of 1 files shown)


colour/models/rgb/transfer_functions/__init__.py
```diff
@@ -777,9 +777,9 @@ def encoding_cctf(value, function='sRGB', **kwargs):
 
     if 'itu-r bt.2100' in function.lower():
         usage_warning(
-            'With the "ITU-R BT.2100" method, only the inverse electro-optical'
+            'With the ITU-R BT.2100 method, only the inverse electro-optical '
```
pavithraes: On adding a space, some checks failed with the following message:

```
E501 line too long (80 > 79 characters)
            'With the "ITU-R BT.2100" method, only the inverse electro-optical '
                                                                               ^
```

To fix this, I removed the double quotes around ITU-R BT.2100 at both places.

KelSolaar: We should keep them ideally for emphasis, is it possible to actually wrap the lines please?

Cheers,

pavithraes: Sure! It's done. Do the current changes work?

KelSolaar: Yes thank you! :)

KelSolaar:
Thanks @pavithraes ! 👍

PR #517, opened 2019-10-03
Move third party notices from LICENSE to THIRD_PARTY

jmertic:
Just a recommendation - this makes the GitHub automatic license identifier properly tag this project as BSD-3-Clause versus Unknown ( which also helps the ASWF Landscape identify the license as well ).


KelSolaar:
Good catch! Thanks @jmertic !

Issue #518, opened 2019-10-04
Replace generator based tests. 


MichaelMauderer:
We have some tests that are based on generators. Their existence is blocking us from moving away from the `nose` test framework. Therefore, they need to go.

PR #519, opened 2019-10-05
PR: Replace generator based tests.

MichaelMauderer:
References #518 .

This is probably the minimum possible change that avoids the generator based tests.

Instead of yielding tests, each check is executed, resulting in a single test that checks the whole dataset for each CAM.

A fancier version could use [pytest's parametrized test fixtures](http://doc.pytest.org/en/latest/parametrize.html) to re-create the one test per case behaviour we had before. But I'm not sure we actually need that.


KelSolaar:
Awesome @MichaelMauderer ! Thanks :) I have considered **pytest** but yeah, let's not go down that road yet!

PR #520, opened 2019-10-09
Fix default values in docstring

skycaptain:
Fixes default values in `colour.models.rgb.ycbcr.YCbCr_to_RGB`s docstring maybe coming from copy-pasting the docstring from `colour.models.rgb.ycbcr.RGB_to_YCbCr`.


KelSolaar:
Thanks Manuel / @skycaptain!

PR #524, opened 2019-10-19
PR: Add code of conduct from website to repository.

MichaelMauderer:
This is mostly because I noticed that GitHub showed the repo as not having a Code of Conduct on the "Insights/Community" site. That is also why it is reformatted as Markdown: from what I have seen, only markdown files are picked up by the system.


KelSolaar:
Sweet thanks! I will copy it to the other repo!

PR #536, opened 2019-12-16
PR: Fix "TypeError" exception caused by "Numpy" 1.18 in "colour.SpectralShape.range" method.

mzszym:
Fix `TypeError: object of type <class 'numpy.float64'> cannot be safely interpreted as an integer` reported with numpy 1.18.0rc1.

### DIFF
--- a/colour/colorimetry/spectrum.py
+++ b/colour/colorimetry/spectrum.py
@@ -37,7 +37,8 @@ from colour.algebra import (Extrapolator, CubicSplineInterpolator,
 from colour.constants import DEFAULT_FLOAT_DTYPE
 from colour.continuous import Signal, MultiSignals
 from colour.utilities import (as_float, first_item, is_iterable, is_numeric,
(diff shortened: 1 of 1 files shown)


KelSolaar:
Hi @mzszym,

Thanks!

Would it be possible to use `colour.utilities.as_int` instead of `int` please?

e.g. 

```python
from colour.utilities import (as_float, as_int, first_item, is_iterable,
                              is_numeric, is_string, is_uniform, interval,
                              runtime_warning)

...

            range_, current_interval = np.linspace(
                self._start, self._end, as_int(samples), retstep=True,
                dtype=dtype)
```

Careful at line length, *flake8* generated errors because of that.

Cheers,

Thomas

mzszym:
Hi @KelSolaar ,

Thank you for the excellent library. Please give a try to the second version of this fix.

Best regards,

Marek

KelSolaar:
Thanks a lot @mzszym and you are welcome :)

Issue #543, opened 2020-01-07
Fix the various "DeprecationWarning" with Python 3.8.


hajimen:
I found two DeprecationWarnings.
`python3.8/site-packages/colour/io/luts/lut.py:20: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working`
`<__array_function__ internals>:5: DeprecationWarning: object of type <class 'numpy.float64'> cannot be safely interpreted as an integer.`
Thanks,

KelSolaar:
Thanks @hajimen! I will take a look. Was it on import?

hajimen:
Dockerfile for Amazon Linux 2:
```
FROM amazonlinux:2
ENV HOME /home/hoge
RUN mkdir $HOME
WORKDIR $HOME
RUN curl -O https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh
RUN bash Miniconda3-latest-Linux-x86_64.sh -b -p ~/miniconda
ENV PATH $HOME/miniconda/bin:$PATH
RUN conda update -n base -c defaults conda
RUN conda create -y -n python380 python=3.8.0
```

bash:
```
source activate python380
pip install numpy==1.17.4  colour-science==0.3.14
```

Python:
```
import unittest
import numpy as np

class TestTarget(unittest.TestCase):
    def test_delta_e_cie2000(self):
        from colour.difference import delta_E_CIE2000
        r = np.arange(0, 100, 20)
        grid = np.array(np.meshgrid(r, r / 2, r / 2)).T.reshape(-1, 3)
        for lab1 in grid:
            for lab2 in grid:
                d = delta_E_CIE2000(lab1, lab2)
```

It generates:
```
/home/hoge/miniconda/envs/python380/lib/python3.8/site-packages/colour/utilities/data_structures.py:31: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working
  from collections import Mapping, MutableMapping
/home/hoge/miniconda/envs/python380/lib/python3.8/site-packages/colour/continuous/signal.py:26: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working
  from collections import Iterator, Mapping, Sequence
[...]

PR #544, opened 2020-01-07
PR: Fix the various "DeprecationWarning" with Python 3.8.

MaxSchambach:
**Numpy 1.18 compatibility fix**
Regarding Issue #543 

I came across this, as the Numpy conversion error causes some of my own unit tests to fail. For me, it actually resulted in an ``TypeError: object of type <class 'numpy.float64'> cannot be safely interpreted as an integer`` unlike the ``DeprecationWarning`` mentioned in #543 .

Numpy's ``TypeError`` fix:
The missing conversion to int is fixed and compatibility to numpy 1.18 restored.
As far as I can tell, there is only one occurence of this issue throughout the colour package.

The ``DeprecationWarning`` regarding abc imports:
To keep backwards compatibility, modules are imported with a try-except clause. Therefore, all code is also compatible with Python>=3.9. I have added one missing try-except clause, but would keep the rest as is to ensure backwards compatibility. Yes, the ``DeprecationWarning``'s are a little annoying, but compatibility is more important, I would say.

### DIFF
--- a/colour/colorimetry/spectrum.py
+++ b/colour/colorimetry/spectrum.py
@@ -469,8 +469,9 @@ class SpectralShape(object):
 
         if self._range is None:
             samples = as_int(
-                round((self._interval + self._end - self._start) /
-                      self._interval))
+                int(
+                    round((self._interval + self._end - self._start) /
+                          self._interval)))
             range_, current_interval = np.linspace(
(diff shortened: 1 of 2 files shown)


KelSolaar:
@MaxSchambach : I will close this one in favour of #544, one of your commits is included.

PR #569, opened 2020-04-12
PR: Fix example of "colour.plotting.plot_RGB_colourspaces_gamuts" definition.

sianyi:
In original example, the argument 'style' is useless.
Correcting the arguments for the example.


KelSolaar:
Good catch @sianyi! Thanks 👍

Issue #571, opened 2020-04-16
PR: Fix incorrect spelling of "HLG" inverse OETF.


jchwei:
HLD => HLG
https://github.com/colour-science/colour/blob/50a40abfdde50b9a583d82de3ea3143fb17a0c35/colour/models/rgb/transfer_functions/__init__.py#L487

KelSolaar:
Thanks and good catch @jchwei!

Would you mind fixing the other instances please?

```diff
diff --git a/README.rst b/README.rst
index 84caa4c6..00e599be 100644
--- a/README.rst
+++ b/README.rst
@@ -808,7 +808,7 @@ OETFs Inverse
 
     >>> sorted(colour.OETF_INVERSES.keys())
     ['ARIB STD-B67',
-     'ITU-R BT.2100 HLD',
+     'ITU-R BT.2100 HLG',
      'ITU-R BT.2100 PQ',
      'ITU-R BT.601',
      'ITU-R BT.709']
diff --git a/colour/models/rgb/transfer_functions/__init__.py b/colour/models/rgb/transfer_functions/__init__.py
index f18635a3..ac35fe5e 100644
--- a/colour/models/rgb/transfer_functions/__init__.py
+++ b/colour/models/rgb/transfer_functions/__init__.py
@@ -484,7 +484,7 @@ def oetf(value, function='ITU-R BT.709', **kwargs):
 
 OETF_INVERSES = CaseInsensitiveMapping({
     'ARIB STD-B67': oetf_inverse_ARIBSTDB67,
-    'ITU-R BT.2100 HLD': oetf_inverse_HLG_BT2100,
+    'ITU-R BT.2100 HLG': oetf_inverse_HLG_BT2100,
     'ITU-R BT.2100 PQ': oetf_inverse_PQ_BT2100,
     'ITU-R BT.601': oetf_inverse_BT601,
     'ITU-R BT.709': oetf_inverse_BT709,
@@ -493,7 +493,7 @@ OETF_INVERSES.__doc__ = """
 Supported inverse opto-electrical transfer functions (OETFs / OECFs).
 
 OETF_INVERSES : CaseInsensitiveMapping
-    **{'ARIB STD-B67', 'ITU-R BT.2100 HLD', 'ITU-R BT.2100 PQ',
+    **{'ARIB STD-B67', 'ITU-R BT.2100 HLG', 'ITU-R BT.2100 PQ',
     'ITU-R BT.601', 'ITU-R BT.709'}**
 """
 
[...]

PR #572, opened 2020-04-16
PR: Fix incorrect spelling of HLG inverse OETF

jchwei:
`'ITU-R BT.2100 HLG'` was incorrectly written as `'ITU-R BT.2100 HLD'` See also #571 .


KelSolaar:
Awesome thanks @jchwei ! :)

Issue #577, opened 2020-04-21
"UnicodeDecodeError" exception raised while installing development version.


sianyi:
Not sure if it is a general issue or just happen at my Windows PC.

I used "pip install -e ." to install developement version in my Windows laptop and it shows 
UnicodeDecodeError: 'cp950' codec can't decode byte 0xf0 in position 2241: illegal multibyte sequence

After made the below change then I can install it properly.
https://github.com/sianyi/colour/commit/12bc73a78a67f31e84b26d38aecae4bf922f0bbc

KelSolaar:
Hi @sianyi,

Which Python version are you using?

Cheers,

Thomas

sianyi:
Python 3.7.3

KelSolaar:
It looks like your locale or something is changing the default Python encoding. Would you mind trying that in the same shell you tried to install Colour:

```python
>>> import sys
>>> sys.getdefaultencoding()
'utf-8'
```

sianyi:
Here is my result..

Python 3.7.3 | packaged by conda-forge | (default, Jul  1 2019, 22:01:29) [MSC v.1900 64 bit (AMD64)] :: Anaconda, Inc. on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import sys
>>> sys.getdefaultencoding()
'utf-8'

KelSolaar:
Ok this is good, can you try that then please:

```python
>>> import locale
>>> locale.getdefaultlocale()
('en_US', 'UTF-8')
```

sianyi:
You catch it!, so it is my enviroement problem ? 

```
>>> import locale
>>> locale.getdefaultlocale()
('en_US', 'cp950')
>>>
```

KelSolaar:
So it is indeed related to your locale!

I think it is reasonable to change the `open` call and force `utf-8` however, I think it would be safer to use an `utf-8` locale generally as this is the default Python encoding and you might run into other issues.

As for the actual change, we should use `codecs.open` because `open` does not support the `encoding` parameter in 2.7 (and we still release for it).

sianyi:
Got it and thanks for the explanation

KelSolaar:
Here is some info on how to change it if you need/want: https://www.tecmint.com/set-system-locales-in-linux/

I will make a note about updating the `open` thing, although because the `setup.py` file is partially generated by Poetry, it is not entirely trivial.

KelSolaar:
[...]

PR #578, opened 2020-04-21
PR: force 'utf-8' encoding in setup.py

sianyi:
Per #577 discussion, if the default encoding is not 'utf-8' in development environment then "UnicodeDecodeError" exception might happen while installing development version.
Force encoding to 'utf-8' for safer.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 from setuptools import setup
+import codecs
 
 package_dir = \
 {'': '.'}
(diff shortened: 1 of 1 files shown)


KelSolaar:
Hi @sianyi,

As I was saying in #577, it is a bit more complicated because the `setup.py` file is generated by Poetry and modified when I build the package: https://github.com/colour-science/colour/blob/develop/tasks.py#L388

sianyi:
Oh, sorry that I didn't catch your point before, looks like only apply the patch can not cover all situation.  I will close this commit.

PR #594, opened 2020-06-01
PR: Micro-optimizations for the "colour.models" sub-package.

OmarWagih1:
Hello, 

These are a few of some micro-optimizations for the models module, the type of these optimizations is a making a variable out of a set of operations that are done more than once and removal of a couple of unnecessary array creations.

Some of these optimizations are shown in other functions (already written) but they're not for these specific functions.

### DIFF
--- a/colour/models/cie_luv.py
+++ b/colour/models/cie_luv.py
@@ -110,10 +110,13 @@ def XYZ_to_Luv(
     with domain_range_scale('100'):
         L = lightness_CIE1976(Y, Y_r)
 
-    u = (13 * L * ((4 * X / (X + 15 * Y + 3 * Z)) -
-                   (4 * X_r / (X_r + 15 * Y_r + 3 * Z_r))))
-    v = (13 * L * ((9 * Y / (X + 15 * Y + 3 * Z)) -
-                   (9 * Y_r / (X_r + 15 * Y_r + 3 * Z_r))))
+    X_Y_Z = X + 15 * Y + 3 * Z
+    X_r_Y_r_Z_r = X_r + 15 * Y_r + 3 * Z_r
+
+    u = (13 * L * ((4 * X / X_Y_Z) -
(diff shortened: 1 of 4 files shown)


KelSolaar:
LGTM! As mentioned on Slack, you should try to get the pre-commit hooks installed: https://www.colour-science.org/contributing/#step-5

Will avoid the sadness of seeing build fails because of linting! :)

colour/models/cie_ucs.py
```diff
@@ -172,7 +172,9 @@ def UCS_to_uv(UVW):
 
     U, V, W = tsplit(to_domain_1(UVW))
 
-    uv = tstack([U / (U + V + W), V / (U + V + W)])
+    U_V_W = U + V + W
+
+    uv = tstack([U / U_V_W, V / U_V_W])
```
MichaelMauderer, line 177: Could the division be done once on `tstack([U,V])`?

colour/models/cie_luv.py
```diff
@@ -110,10 +110,13 @@ def XYZ_to_Luv(
     with domain_range_scale('100'):
         L = lightness_CIE1976(Y, Y_r)
 
-    u = (13 * L * ((4 * X / (X + 15 * Y + 3 * Z)) -
-                   (4 * X_r / (X_r + 15 * Y_r + 3 * Z_r))))
-    v = (13 * L * ((9 * Y / (X + 15 * Y + 3 * Z)) -
-                   (9 * Y_r / (X_r + 15 * Y_r + 3 * Z_r))))
+    X_Y_Z = X + 15 * Y + 3 * Z
+    X_r_Y_r_Z_r = X_r + 15 * Y_r + 3 * Z_r
```
MichaelMauderer, line 114: Does this term have some actual meaning behind it that could serve as the name?
(Same for `X_Y_Z` further down)

PR #599, opened 2020-06-12
PR: Micro-optimizations for the "colour.models" sub-package.

OmarWagih1:
Hello, This is a pull request for small microoptimizations. 

The optimizations for the models/rgb are small and get about 5-10% speedup, while the optimizations for the transfer_functions module although small get about 40-50% for the last two and about 90% for the aces function.

As for comments on the previous pull request (sorry my local repository got deleted and so i had to close that one)

The X_Y_Z and X_r_Y_r_Z_r variables don't seem to have a more descriptive name as far as I've searched and the U_V_W would need to be tstacked to divide over the entire variable, losing some performance.

### DIFF
--- a/colour/models/cie_luv.py
+++ b/colour/models/cie_luv.py
@@ -110,10 +110,11 @@ def XYZ_to_Luv(
     with domain_range_scale('100'):
         L = lightness_CIE1976(Y, Y_r)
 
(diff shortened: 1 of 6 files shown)


KelSolaar:
Thanks @OmarWagih1, merged!

PR #601, opened 2020-06-21
PR: Micro-optimizations for the "colour.temperature" sub-package.

OmarWagih1:
These are microoptimizations to the Temperature module.

These optimizations are relatively small but the time decrease is 10-50% depending on the function, which is quite substantial considering these functions are used in the objective function (and so called hundreds or thousands of times) in the optimization problem of each of their respective counter functions. 

This marks the end of my benchmarking and microoptimizations phase.


KelSolaar:
Excellent! The builds are breaking because of Numpy 1.19, should be fixed soon, you will need to rebase on top of develop.

PS: Is it possible to use nice commit messages please: https://www.colour-science.org/contributing/#commits

OmarWagih1:
@KelSolaar This build should pass, there's something wrong in installing packages in MacOS which I'm not sure what the cause of is. Also, Let me know if this commit message is more suitable! Thanks!

KelSolaar:
LGTM! Thanks @OmarWagih1!

PR #625, opened 2020-08-28
PR: CuPy Based GPU Backend

OmarWagih1:
The state of the repository right now is that most tests pass (only about 30 tests fail) with a default backend of CuPy.

All failed tests are mentioned in this document with the reason of their failure

https://docs.google.com/document/d/1Ag7Yh6Ovi1zKwZSv2p-ckwC6gTuZrzb5CUUt3f7iShc/edit

Some tests fail because the randomization between CuPy and NumPy is different, so while the code works correctly the tests do not pass. Some functions are not implemented with CuPy backend and others are not compatible with CuPy.

The code needs working on to be better looking as some Numpy/Cupy switches were crudely done.


KelSolaar:
Thanks @OmarWagih1!

Would it be possible to rebase this on develop so that there are only the commits pertaining to the GPU backend please?

PR #651, opened 2020-11-12
PR: Fix "CCT_to_xy_McCamy1992" comment typo.

selfshadow:
I believe the comment should say "*not* a bijective function", in order to match the usage warning later in the same file.


KelSolaar:
Ah! Good catch Stephen! It seems like it is the week where we fixed typos in the docs :)

Thanks.

selfshadow:
Happy to help! I'll keep an eye out for others in the future.

KelSolaar:
Thank you!

PR #657, opened 2020-12-03
PR: The Great Python2Ectomy!

KelSolaar:
This PR removes Python 2.7 from the codebase.

Here are a few other things to do down the line and as we go:

- Using `f-strings`
- Using `pathlib` instead of `os.path`
- Remove `OrderedDict` usage as Python 3 dictionary are ordered by default


KelSolaar:
We will have to do a coverage pass post-merge but it seems like I managed to get all the references out, took a bit longer than expected.

KelSolaar:
And merging! 🎊

Issue #486, opened 2019-07-02
Implement support for Y'CbCr matrices generation.


KelSolaar:
@nick-shaw and I talked about implementing a new `colour.RGB_to_YCbCr_matrix` definition that would generate the matrices used by the various *Y'CbCr* flavours, it would eventually be used by the current `colour.RGB_to_YCbCr` and `colour.YCbCr_to_RGB` definitions.

It would be beneficial for generating the matrices for usage in other applications and also probably faster.

KelSolaar:
Adding some related code for future reference, it misses the offset though:

```python
def compute_YCbCr_matrix(weights, is_legal=False, bit_depth=8):
    Kr, Kb = weights
    Y_min, Y_max, C_min, C_max = colour.models.rgb.ycbcr.YCbCr_ranges(bit_depth, is_legal, False)
    Y = np.array([Kr, (1 - Kr - Kb), Kb])
    Cb = 0.5 * (np.array([0, 0, 1]) - Y) / (1 - Kb)
    Cr = 0.5 * (np.array([1, 0, 0]) - Y) / (1 - Kr)
    Y *= Y_max - Y_min
    Cb *= C_max - C_min
    Cr *= C_max - C_min
    return np.linalg.inv(np.vstack([Y, Cb, Cr]))
```

nick-shaw:
Should that code not also include the `is_int` parameter?

Would it be useful also to create a corresponding `compute_YCbCr_offsets()` function, even if that was relatively trivial? It would presumably be something like:
```python
def compute_YCbCr_offsets(weights, is_legal=False, bit_depth=8, is_int=False):
    Y_min, Y_max, C_min, C_max = colour.models.rgb.ycbcr.YCbCr_ranges(bit_depth, is_legal, is_int)
    Y_offset = Y_min
    C_offset = (C_min + C_max) / 2
    return np.array([Y_min, C_offset, C_offset]
```
Then you could calculate the matrix and offsets and do:
```python
 YCbCr = colour.utilities.dot_vector(matrix, RGB) + offsets
```

KelSolaar:
Absolutely! It is not complete at all, I just pasted it so that I don't have to scroll for 10 hours in Slack :)

remia:
> ```python
>  YCbCr = colour.utilities.dot_vector(matrix, RGB) + offsets
> ```
Maybe propose an option to have the offset right into a 3x4 or 4x4 matrix as well ?

KelSolaar:
[...]

PR #664, opened 2020-12-23
PR: Implement support for Y'CbCr matrices generation.

KelSolaar:
References #486.

### DIFF
--- a/colour/__init__.py
+++ b/colour/__init__.py
@@ -127,9 +127,10 @@ from .models import (
     chromatically_adapted_primaries, eotf, eotf_inverse, full_to_legal,
     gamma_function, hdr_CIELab_to_XYZ, hdr_IPT_to_XYZ, legal_to_full,
     linear_function, log_decoding, log_encoding, matrix_RGB_to_RGB,
-    normalised_primary_matrix, oetf, oetf_inverse, ootf, ootf_inverse,
-    primaries_whitepoint, sRGB_to_XYZ, uv_to_Luv, uv_to_UCS, xyY_to_XYZ,
-    xyY_to_xy, xy_to_Luv_uv, xy_to_UCS_uv, xy_to_XYZ, xy_to_xyY)
+    matrix_YCbCr, normalised_primary_matrix, oetf, oetf_inverse, offset_YCbCr,
+    ootf, ootf_inverse, primaries_whitepoint, sRGB_to_XYZ, uv_to_Luv,
+    uv_to_UCS, xyY_to_XYZ, xyY_to_xy, xy_to_Luv_uv, xy_to_UCS_uv, xy_to_XYZ,
+    xy_to_xyY)
 from .corresponding import (
     BRENEMAN_EXPERIMENTS, BRENEMAN_EXPERIMENT_PRIMARIES_CHROMATICITIES,
     CORRESPONDING_CHROMATICITIES_PREDICTION_MODELS, CorrespondingColourDataset,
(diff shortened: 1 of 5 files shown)


KelSolaar:
@nick-shaw : Might be worth one of your eyes! :)

ValZapod:
Couple of points and/or references: in [Supplement 18](https://www.itu.int/itu-t/recommendations/rec.aspx?id=13441&lang=en) of ITU-T (page 24) a matrix with 14 digits after decimal point is given and is to be used for Y′CbCr to R′G′B′.

The second point is that I recommend doing immediately xvYCC colorspace. That is the negative values from XYZ to RGB (linear RGB that this) are further defined while the 601 and 709 stuff stays the same. That is the standard nowadays. Used in Kodak Photo CD and is used now very often in HDMI.
I will attach the latest version of the standard (with all errata), the matrix from XYZ to RGB and non-linearity transfer function (just a continuation of the very same BT.709 and BT.601 transfer function) are in 5.3. 

I will also point out that "for TransferCharacteristics equal to 1, 6, 14 or 15, α has the value 1 + 5.5 * β = 1.099296826809442... and β has the value 0.018053968510807...." and 0.099 value is just 1 - α. That is from H.273.

nick-shaw:
> Couple of points and/or references: in [Supplement 18](https://www.itu.int/itu-t/recommendations/rec.aspx?id=13441&lang=en) of ITU-T (page 24) a matrix with 14 digits after decimal point is given and is to be used for Y′CbCr to R′G′B′.

This function would be able to calculate those 14 decimal place coefficients (or higher precision if desired). The 14 decimal place coefficients are just the 4 decimal place constants from BT.2020 run through the weighting to matrix calculations.

nick-shaw:
```
>>> np.set_printoptions(suppress=True, formatter={'float_kind':'{:0.14f}'.format})
>>> matrix_YCbCr(K=WEIGHTS_YCBCR['ITU-R BT.2020'])
array([[1.00000000000000, -0.00000000000000, 1.47460000000000],
       [1.00000000000000, -0.16455312684366, -0.57135312684366],
       [1.00000000000000, 1.88140000000000, 0.00000000000000]])
```
This is exactly the matrix shown in the document.

nick-shaw (commented):
Looks good to me, except perhaps the docstring should be more explicit that the returned matrix is the Y'CbCr to R'G'B' matrix, not the R'G'B' to Y'CbCr matrix. Other than that I am happy to approve the PR.

KevinJW:
I haven't fully looked at this, but I wonder if [ITU T-REC-H.273](https://www.itu.int/rec/T-REC-H.273-201612-I/en) might have something to say on these, at least it states the expected behaviours for a number of the combinations. It does appear to be attempting to be definitive: `This Specification can be used to provide universal descriptions to assist interpretation of video and image signals following decoding or to describe the properties of these signals before they are encoded`

Kevin

nick-shaw:
Not sure I completely follow, @KevinJW. Are you suggesting we adopt the  "Video Code Points" values as an (optional?) way of passing the `K` and `is_legal` arguments to the functions? This feels like it is more a part of the metadata layer that we do not yet have. Ideally the functions would also return metadata describing how the result was encoded.

ValZapod:
> adopt the "Video Code Points" values as an

That is a bad idea, unless you wanna attach ffmpeg to yourself and become a new super video player... You can, BTW, considering ;)

> This is exactly the matrix shown in the document.

Yeah, nice. LGTM, then! BTW, can you add Kodak Photocd quantization? See https://github.com/DavidGriffith/finx/blob/17825bf28b2351f96867b34a7dc39f7997e2f9ff/nx-X11/programs/Xserver/XIE/mixie/process/mprgb.c#L1004 and this doc http://scarse.sourceforge.net/docs/kodak/pcd-045.pdf

> 1.099296826809442... and β has the value 0.018053968510807

This is BTW used in ffmpeg, just FYI.
https://github.com/FFmpeg/FFmpeg/blob/069d2b4a50a6eb2f925f36884e6b9bd9a1e54670/libavutil/color_utils.c#L58

While you use less precise https://github.com/colour-science/colour/search?q=0.018

nick-shaw:
> BTW, can you add Kodak Photocd quantization?

The `RGB_to_YCbCr` function can already use arbitrary output scaling and offsets, using the `out_range` argument. See [here](https://colour.readthedocs.io/en/latest/generated/colour.RGB_to_YCbCr.html?highlight=out_range) for details. So the PhotoCD quantisation could be added using that together with the `out_int` argument.

KelSolaar:
Thanks for the review @nick-shaw! I addressed the docstring notes and rebased on-top of develop. I will merge after the tests pass.

KevinJW:
The numbers in that line in FFmpeg would be my "fault", can't remember exactly what source I used for it

Kevin

ValZapod:
> FFmpeg would be my "fault", can't remember exactly what source I used for it

H.273, of course! Or H.264, as it was there before that.

Issue #665, opened 2020-12-23
Implement support for "Oklab" colourspace.


KelSolaar:
References
---
- https://bottosson.github.io/posts/oklab/

PR #667, opened 2020-12-24
PR: Implement support for "Oklab" colourspace.

KelSolaar:
References #665.


KelSolaar:
It is pretty much like IPT & co. Merging!

Issue #301, opened 2016-12-22
Implement support for "Dolby ICTCP" colourspace.


KelSolaar:
References
-----------

- http://wwwen.zte.com.cn/endata/magazine/ztecommunications/2016/1/articles/201603/P020160311294995790791.pdf

ValZapod:
BT.2100-2 (that is version 2) created HLG ICtCp matrix. I also updated Wikipedia for that new matrix https://en.wikipedia.org/wiki/ICtCp#Derivation (that is last matrix "in HLG:"). I will also recommend to check that all 3 (two for PQ and one new for HLG) inverted matrixes are correct to 15 decimal places as recommended by Series H Supplement 18 [*] https://www.itu.int/rec/dologin_pub.asp?lang=e&id=T-REC-H.Sup18-201710-I!!PDF-E&type=items

[*] But you can just use WolframAlpha to get the very same matrix for PQ https://www.wolframalpha.com/input/?i=Inverse%5BSetPrecision%5B%7B%7B2048%2F4096.%2C+2048%2F4096.%2C+0%7D%2C+%7B6610%2F4096.%2C+-13613%2F4096.%2C7003%2F4096.%7D%2C+%7B17933%2F4096.%2C+-17390%2F4096.%2C+-543%2F4096.%7D%7D%2C+30%5D%5D 
or even matrix with absolute precision in integers
https://www.wolframalpha.com/input/?i=Inverse%5BSetPrecision%5B%7B%7B2048%2F4096.%2C+2048%2F4096.%2C+0%7D%2C+%7B6610%2F4096.%2C+-13613%2F4096.%2C7003%2F4096.%7D%2C+%7B17933%2F4096.%2C+-17390%2F4096.%2C+-543%2F4096.%7D%7D%2C+Inf%5D%5D

Also the module of values in 2x2 submatrix of that inverted matrix are the same if you did not recognize that. Very cool, right?

PR #669, opened 2020-12-28
PR: Implement wrapper "colour.XYZ_to_ICTCP" and "colour.ICTCP_to_XYZ" definitions.

KelSolaar:
This PR implements two convenient definitions to work with the `IC_TC_P` colourspace:

- `colour.XYZ_to_ICTCP`
-  `colour.ICTCP_to_XYZ`

Both expose the source illuminant and chromatic adaptation transform as arguments so that the incoming data can be converted from/to D65 as used by ITU-R BT.2020.


ValZapod:
ICtCp matrix (from LMS to ICtCp) for HLG changed in BT.2100-2. Also see my comment in #301. That is what was forgotten in #450.

KelSolaar:
Thanks @ValZapod! We will look at that!

PR #660, opened 2020-12-12
PR: Implement support  for "DaVinci Intermediate" log encoding.

fredsavoir:

PR #670, opened 2020-12-31
PR: Implement support for "DaVinci Intermediate" log encoding.

KelSolaar:
Replaces #660 as @fredsavoir deleted the branch.

We might actually merge it "ad interim" now that the commits are squashed and do not affect the gamut primaries. Still requires various documentation updates, etc... before though.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -968,6 +968,7 @@ OETFs
 
     >>> sorted(colour.OETFS)
     ['ARIB STD-B67',
+     'DaVinci Intermediate',
      'ITU-R BT.2100 HLG',
      'ITU-R BT.2100 PQ',
      'ITU-R BT.601',
--- a/colour/models/rgb/datasets/davinci_wide_gamut.py
+++ b/colour/models/rgb/datasets/davinci_wide_gamut.py
@@ -11,12 +11,17 @@ References
 ----------
 -   :cite:`BlackmagicDesign2020` : Blackmagic Design. (2020).
     DaVinci Wide Gamut - DaVinci Resolve Studio 17 Public Beta 1.
+-   :cite:`BlackmagicDesign2020a` : Blackmagic Design. (2020). Wide Gamut
+    Intermediate DaVinci Resolve. Retrieved December 12, 2020, from
+    https://documents.blackmagicdesign.com/InformationNotes/\
+DaVinci_Resolve_17_Wide_Gamut_Intermediate.pdf?_v=1607414410000
 """
 
 import numpy as np
 
 from colour.colorimetry import CCS_ILLUMINANTS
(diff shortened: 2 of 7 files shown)


KelSolaar:
I think this is ready for review, quite a bit of code formatting and consistency improvements. Worth noting that the functions are OETF / OETF-1 and not log encoding curves.

zachlewis:
> Worth noting that the functions are OETF / OETF-1 and not log encoding curves.

Just to be crystal clear, these should be reclassified as log encoding / decoding functions, and shouldn't remain OETF / OETF-1, correct?

KelSolaar:
No the reverse, they were classified as log encoding but Blackmagic states it is an OETF and its inverse, so I renamed Fred's definitions to reflect that.

nick-shaw:
How do we define if something is an OETF or a log encoding? Is it just based on what the published documentation calls it? A log encoding **is** one form of OETF, isn't it? In general we have OETFs for those where the input is in the 0-1 range, and log encoding for scene-referred HDR, don't we? The BMD paper is a bit confusing, because it refers to nits as if it is an inverse EOTF, but Resolve Colour management clearly treats it as a scene referred working space, applying tone and gamut mapping by default when converting to display referred.

KelSolaar:
> How do we define if something is an OETF or a log encoding? Is it just based on what the published documentation calls it? A log encoding is one form of OETF, isn't it?

I would say that if the vendor explicitly states it is an OETF, we should not call it something else because it is what less tech savvy people will refer to and even though it might be incorrect, it will only serve confusion! Should BMD change their "naming convention", we could/should change it accordingly.

fredsavoir:
Fully agree .

Thanks

Fred

nick-shaw (commented):
Looks good to me. My only comment would be that there are redundant brackets in `davinci_intermediate.py` line 164. `(2 ** ((V / DI_C) - DI_B)) - DI_A` could be written `2 ** ((V / DI_C) - DI_B) - DI_A`.
And do we not bother including XYZ matrices from the white paper in the colour space definition if the ones derived from the primaries match to the number of decimal places given in the paper?

KelSolaar:
Thanks @nick-shaw!

> `(2 ** ((V / DI_C) - DI_B)) - DI_A` could be written `2 ** ((V / DI_C) - DI_B) - DI_A`.

Ah! I think it might have come when Fred copied the reference implementation from the Whitepaper as it used the `power` function. I have just removed them!

Interestingly enough, I got used to PyCharm hinting when they are redundant but it seems like here it does not work, maybe broken functionality in my current version.

> And do we not bother including XYZ matrices from the white paper in the colour space definition if the ones derived from the primaries match to the number of decimal places given in the paper?

I will add them in a separate commit!

PR #674, opened 2021-01-13
PR: Red Log3G10 Linear Extension

### LINKED ISSUE #673: Red Log3G10 Missing Linear Extension
Greetings!
I noticed that in `colour.models.rgb.transfer_functions.red_log`, `log_encoding_Log3G10_v2` does not include the linear extension included in the [official whitepaper](https://docs.red.com/955-0187/PDF/915-0187%20Rev-C%20%20%20RED%20OPS,%20White%20Paper%20on%20REDWideGamutRGB%20and%20Log3G10.pdf). 

Is this an issue, perhaps because this implementation predated the whitepaper's release? Or did the v2 method exist for a while in RedCineX Pro before the whitepaper was released? Should there be a v3 matching the whitepaper?

I'd be happy to put in a pull request if any knowledgeable folks here know the right solution.

Thanks!

jedypod:
Fixes #673.

- Add `log_encoding_Log3G10_v3` and `log_decoding_Log3G10_v3`. 
- Switch default Log3G10 transform to v3.
- Add citation for Log3G10 Whitepaper.

Be gentle with me, it's my first pull request for colour :)


jedypod:
D: 
That's a lot of failed checks. I did run `poetry run invoke tests` on my local machine and there no failures in the red_log module... But it looks like some of the unit tests return slightly different values on other platforms...

KelSolaar:
Thanks @jedypod! Looks good overall

I think it breaks in most of the doctests because Numpy returns a 0d array, when this is the case, you can do a cast with `colour.utilities.as_numeric` before the return.

Haven't looked why but this one seems broken though: https://github.com/colour-science/colour/pull/674/checks?check_run_id=1696289735#step:11:575

We will need unit tests here also: https://github.com/KelSolaar/colour/blob/f4040f1e4d17a326ed90492c2b51b4243ef4c64b/colour/models/rgb/transfer_functions/tests/test_red_log.py

jedypod:
Pushed some more changes: 
- Added unit tests for *_v3 variant in `test_red_log` - (Sorry for missing that! Starting to get more familiar with the code :) )
- Add np.abs() [here](https://github.com/jedypod/colour/blob/df712dea326a819fa88e90e747f2ff61a77ac689/colour/models/rgb/transfer_functions/red_log.py#L490) : despite the np.where, np.log10 was unhappy with x being negative.
- Add as_numeric call on the return value of the *_v3 variants only. (Wasn't sure if it was appropriate for the other functions). 

> Haven't looked why but this one seems broken though: https://github.com/colour-science/colour/pull/674/checks?check_run_id=1696289735#step:11:575

- I wasn't able to figure out this error... Running `print(colour.models.rgb.transfer_functions.red_log.log_encoding_Log3G10_v3(-0.1))` results in `-1.367343`  as expected, but when running the tests it always errored saying the result was 0.0. So I took the lazy route and just removed the negative number doctest strings, since the unit tests should take care of it?

Let me know if I missed anything else!

KelSolaar:
Thx again! Minor things:

> * Add np.abs() [here](https://github.com/jedypod/colour/blob/df712dea326a819fa88e90e747f2ff61a77ac689/colour/models/rgb/transfer_functions/red_log.py#L490) : despite the np.where, np.log10 was unhappy with x being negative.

We might actually want multiply by the sign of x outside of the log, e.g.

> np.sign(x) * a * np.log10((np.abs(x) * b) + 1.0)

And conversely, for the inverse:

> np.sign(y) * (10 ** (np.abs(y) / a) - 1.0) / b - c

https://www.desmos.com/calculator/ss33wrvr03

Here is the citation key / bibliography btw:

> REDDigitalCinema2017

> RED Digital Cinema. (2017). White Paper on REDWideGamutRGB and Log3G10. https://www.red.com/download/white-paper-on-redwidegamutrgb-and-log3g10

jedypod:
Thanks @KelSolaar! I was totally shooting in the dark with the citation, so I appreciate the help. 

I've pushed the requested changes to my branch.

KelSolaar:
Cheers and thanks for your first contribution @jedypod, merged!

jedypod:
:D 
Thank you!

PR #756, opened 2021-01-17
PR: Implement support for "ICtCp" matrix as given in "ITU-R BT.2100-2" for "HLG".

KelSolaar:
This PR implements the *ICtCp* matrix as given in *ITU-R BT.2100-2* for *HLG*.

**ITU-R BT.2100-1**

image

**ITU-R BT.2100-2**

image

Thanks @Valzapod for the heads-up!

### DIFF
--- a/colour/models/rgb/ictcp.py
+++ b/colour/models/rgb/ictcp.py
@@ -14,6 +14,12 @@ References
 ----------
 -   :cite:`Dolby2016a` : Dolby. (2016). WHAT IS ICtCp? - INTRODUCTION.
     https://www.dolby.com/us/en/technologies/dolby-vision/ICtCp-white-paper.pdf
+-   :cite:`InternationalTelecommunicationUnion2018` : International
+    Telecommunication Union. (2018). Recommendation ITU-R BT.2100-2 - Image
+    parameter values for high dynamic range television for use in production
+    and international programme exchange.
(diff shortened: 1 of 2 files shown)


ValZapod:
The only thing that needs to be checked is that the inverted matrix for LMS --> ICTCP for HLG is as follows (per ITU-T H. Supplement 18 you need only first 15 digits after decimal point):

https://www.wolframalpha.com/input/?i=Inverse%5BSetPrecision%5B%7B%7B2048%2F4096.%2C+2048%2F4096.%2C+0%7D%2C+%7B3625%2F4096.%2C+-7465%2F4096.%2C3840%2F4096.%7D%2C+%7B9500%2F4096.%2C+-9212%2F4096.%2C+-288%2F4096.%7D%7D%2C+30%5D%5D

KelSolaar:
The inverted matrix is computed at machine precision, so double, i.e. float64 and should be as accurate as needed.

nick-shaw (commented):
Looks good to me. I don't suppose there is any benefit in allowing `method='ITU-R BT.2100-2 PQ'` as an alias to `Dolby 2016`?

KelSolaar:
Thanks @nick-shaw!

For the alias, I thought about it and decided it was not worth cluttering the API but given you mentioned it, I reckon there is virtue having it from if not consistency at least a documentation standpoint, so I have added it!

PR #763, opened 2021-01-24
PR: Add "DaVinci Wide Gamut" conversion matrices as given in the white paper.

KelSolaar:
This PR adds the *DaVinci Wide Gamut* conversion matrices as given in the white paper and in support to https://github.com/colour-science/colour/discussions/762.


zachlewis (approved):
Confirmed, these values exactly match those in the cited reference 👍

fredsavoir:
Hi, why we have use the value from
Paper and not compute it ourself.

Thanks

KelSolaar:
> Hi, why we have use the value from
> Paper and not compute it ourself.
> 
> Thanks

We don't! See the conversation here: https://github.com/colour-science/colour/discussions/762

KelSolaar:
Merging this one, the current default irrespective of #762 is to use the derived NPM.

PR #772, opened 2021-02-03
PR: Implement support for "Huang et al. (2015)" Power-Functions.

KelSolaar:
Long time coming! This PR implements support for *Huang et al. (2015)* Power-Functions. They basically improve the colour difference formulas by applying a simple power-function on the computed delta_E value:

- `colour.difference.power_function_Huang2015`

image

I also took that opportunity to implement the STRESS index as given by García et al. (2007):

- `colour.index_stress`, maybe we should call it `stress_index` but it is consistent right now.
- `colour.INDEX_STRESS_METHODS`
- `colour.difference.index_stress_Garcia2017`

image


KelSolaar:
Been a few weeks, merging this one!

PR #793, opened 2021-03-12
PR: Implement support for "Blackmagic Design" camera colour encodings.

nick-shaw:
Add basic support for Blackmagic camera colour encodings.


KelSolaar:
Moving some of the discussion here so that they don't disappear in Slack's Void.

So we have three notebooks doing curve fitting:

- @nick-shaw : https://colab.research.google.com/drive/1XNbxO6O6bGjtC1Igzpo7gDE4dtlHPeyW?usp=sharing

This one is the original Nick built with semi-manual break-point and fitting, it assumes a quasi-log function, i.e. `colour.models.logarithmic_function_quasilog` definition and a linear function.

- @KevinJW  : https://colab.research.google.com/drive/1MXUAwuI-9O3tvdvx5NTmGkZtdCGky2kw?usp=sharing

Based on Nick's notebook and implementing fully automatic fitting, it also assumes a quasi-log function, i.e. `colour.models.logarithmic_function_quasilog` definition and a linear function.

- @KelSolaar  : https://colab.research.google.com/drive/1tIdZM9zmcXYEd4IHHJQSgc0lY8PpGGAI?usp=sharing

Based on Nick's notebook and implementing fully automatic fitting, it also assumes the piece-wise CLF camera log function, i.e. `colour.models.logarithmic_function_camera`.

I would be keen to have a single notebook we can refer to but besides that, we should decide on how we want to implement the functions in `Colour`. I was hoping that we could use `colour.models.logarithmic_function_camera` definition, but the last two LUTs, i.e. *BMD_POCKET_4K_FILM_LUT* and *BMD_POCKET_6K_FILM_LUT*, have a joining problem which is evident in my notebook when using `colour.models.logarithmic_function_camera`:

image

It can be corrected with a blending function but at this point, we might as well use `colour.models.logarithmic_function_quasilog` definition and a linear segment:

image

Thoughts?

PR #840, opened 2021-07-31
PR: Implement support for spectral data lazy loading and reshape cache.

KelSolaar:
This PR implements support for lazy loading of spectral data so that importing Colour is significantly faster, i.e. ~= 1.5sec vs 
~5.5sec for 0.3.16 on my old Macbook Pro 2014.

Note: Some of the import improvements are also pertaining to some changes I made in the `colour.plotting.tm3018` module.

image

image

The shipped spectral data is not instantiated until first used after full package import. The implementation uses the `colour.utilities.LazyCaseInsensitiveMapping` class that initially stores callables that are replaced by the data when an item is being requested:

```
    def __getitem__(self, item):
        import colour

        value = super(LazyCaseInsensitiveMapping, self).__getitem__(item)

        if callable(value) and hasattr(colour, '__disable_lazy_load__'):
            value = value()
            super(LazyCaseInsensitiveMapping, self).__setitem__(item, value)

        return value
```

A typical 0.3.16 dataset is as follows:

```python
MSDS_CAMERA_SENSITIVITIES_DSLR = CaseInsensitiveMapping({
    'Nikon 5100 (NPL)':
        RGB_CameraSensitivities(
            DATA_CAMERA_SENSITIVITIES_DSLR['Nikon 5100 (NPL)'],
            name='Nikon 5100 (NPL)'),
    'Sigma SDMerill (NPL)':
        RGB_CameraSensitivities(
            DATA_CAMERA_SENSITIVITIES_DSLR['Sigma SDMerill (NPL)'],
            name='Sigma SDMerill (NPL)')
})
```

it is now defined as follows:

```python
MSDS_CAMERA_SENSITIVITIES_DSLR = LazyCaseInsensitiveMapping({
    'Nikon 5100 (NPL)':
        partial(
            RGB_CameraSensitivities,
            DATA_CAMERA_SENSITIVITIES_DSLR['Nikon 5100 (NPL)'],
            name='Nikon 5100 (NPL)'),
    'Sigma SDMerill (NPL)':
        partial(
            RGB_CameraSensitivities,
            DATA_CAMERA_SENSITIVITIES_DSLR['Sigma SDMerill (NPL)'],
            name='Sigma SDMerill (NPL)')
})
```

In order to be able to compose datasets during import and avoid instantiation
to spectral data, the process is prevented by the `colour.__disable_lazy_load__` attribute that is only set after the package is fully imported at the very end of `colour.__init__`.

This new mechanism also means that we have to update the signature of all the
objects that use a default spectral distribution or multi-spectral distribution
as a default argument otherwise, they are defined with the `partial` as a default argument which is unsuitable.

It is worth noting that this does not affect packages using `Colour` because it is `Colour` import time "magic" only.

I took that opportunity to introduce high-level wrappers, i.e. `colour.colorimetry.reshape_sd` and `colour.colorimetry.reshape_msds`, that handle alignement, extrapolation, interpolation and trimming while avoiding to have to make a copy systematically. The other benefit is that we can also cache the operation, and
a subsequent call for the same spectral data with exact same reshaping arguments will use the cached spectral data instead of regenerating, thus better performance.

A typical 0.3.16 definition is as follows:

```python
def sd_to_XYZ_integration(
        sd,
        cmfs=MSDS_CMFS_STANDARD_OBSERVER['CIE 1931 2 Degree Standard Observer']
        .copy().trim(SPECTRAL_SHAPE_DEFAULT),
        illuminant=sd_ones(),
        k=None):

    if illuminant.shape != cmfs.shape:
        runtime_warning(
            'Aligning "{0}" illuminant shape to "{1}" colour matching '
            'functions shape.'.format(illuminant.name, cmfs.name))
        illuminant = illuminant.copy().align(cmfs.shape)

    if sd.shape != cmfs.shape:
        runtime_warning('Aligning "{0}" spectral distribution shape to "{1}" '
                        'colour matching functions shape.'.format(
                            sd.name, cmfs.name))
        sd = sd.copy().align(cmfs.shape)
```

It is now as follows:

```python
def sd_to_XYZ_integration(sd, cmfs=None, illuminant=None, k=None):
    if cmfs is None:
        cmfs = reshape_msds(MSDS_CMFS_STANDARD_OBSERVER[_DEFAULT_MSDS_CMFS],
                            SPECTRAL_SHAPE_DEFAULT)

    if illuminant is None:
        illuminant = sd_ones()

    if illuminant.shape != cmfs.shape:
        runtime_warning(
            'Aligning "{0}" illuminant shape to "{1}" colour matching '
            'functions shape.'.format(illuminant.name, cmfs.name))
        illuminant = reshape_sd(illuminant, cmfs.shape)

    if sd.shape != cmfs.shape:
        runtime_warning('Aligning "{0}" spectral distribution shape to "{1}" '
                        'colour matching functions shape.'.format(
                            sd.name, cmfs.name))
        sd = reshape_sd(sd, cmfs.shape)
```

### DIFF
--- a/colour/colorimetry/spectrum.py
+++ b/colour/colorimetry/spectrum.py
@@ -11,6 +11,8 @@ Defines the classes and objects handling spectral data computations:
 -   :class:`colour.MultiSpectralDistributions`
 -   :func:`colour.colorimetry.sds_and_msds_to_sds`
 -   :func:`colour.colorimetry.sds_and_msds_to_msds`
+-   :func:`colour.colorimetry.reshape_sd`
+-   :func:`colour.colorimetry.reshape_msds`
 
 References
 ----------
@@ -25,14 +27,16 @@ References
 """
 
 import numpy as np
+from collections.abc import Mapping
 
 from colour.algebra import (Extrapolator, CubicSplineInterpolator,
                             SpragueInterpolator)
 from colour.constants import DEFAULT_FLOAT_DTYPE
 from colour.continuous import Signal, MultiSignals
-from colour.utilities import (as_float, as_int, first_item, is_iterable,
-                              is_numeric, is_string, is_uniform, interval,
-                              runtime_warning, tstack)
+from colour.utilities import (as_float, as_int, copy_definition, filter_kwargs,
+                              first_item, is_iterable, is_numeric, is_string,
+                              is_uniform, interval, runtime_warning, tstack,
+                              validate_method)
 
 __author__ = 'Colour Developers'
 __copyright__ = 'Copyright (C) 2013-2021 - Colour Developers'
@@ -43,7 +47,8 @@ __status__ = 'Production'
 
 __all__ = [
     'SpectralShape', 'SPECTRAL_SHAPE_DEFAULT', 'SpectralDistribution',
-    'MultiSpectralDistributions', 'sds_and_msds_to_sds', 'sds_and_msds_to_msds'
+    'MultiSpectralDistributions', 'sds_and_msds_to_sds',
+    'sds_and_msds_to_msds', 'reshape_sd', 'reshape_msds'
 ]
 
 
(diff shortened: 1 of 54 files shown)


colour/colorimetry/spectrum.py
```diff
@@ -2659,3 +2664,75 @@ def sds_and_msds_to_msds(sds):
             tstack(values), shape.range(), labels, strict_labels=strict_labels)
 
     return msds
+
+
+_CACHE_RESHAPED_SDS_AND_MSDS = {}
+
+
+def reshape_sd(sd, shape=SPECTRAL_SHAPE_DEFAULT, method='Align', **kwargs):
```
MichaelMauderer, line 2672: Could this also be achieved via an [LRU cache](https://docs.python.org/3/library/functools.html#functools.lru_cache), instead of a custom caching implementation? Just adding the annotation seems simpler and less error-prone.

KelSolaar: This is a good question, right now the cache is a simple as it can be, i.e. a dict, because I did not felt like it warranted anything fancier than that. However, with #846, it should be trivial to play and test various style of caches!

KelSolaar: Oh and as far as using an LRU cache via the decorator, it would not work because so of the arguments are dictionaries, e.g. `extrapolotor_kwargs={'left': 0, 'method': constant}`, and thus they are not hashable (and usable directly)!

KelSolaar: So there is that option: https://stackoverflow.com/a/53394430/931625 but it is becoming overkill.

We also don't have too much control over the cache like I would like to have with #846.

Another minor thing but also worth noting is that keyword arguments order is important for `lru_cache`, e.g. `foo(a=1, b=2)` does not have the same cache entry as `foo(b=2, a=1)`.

KelSolaar: I'm going to merge the PR, it is becoming a blocker and given we cannot use the decorator directly because of the `dict` args this point can be resolved.

PR #845, opened 2021-08-12
PR: Implement support for DIN99 variants.

SGeetansh:


colour/models/din99.py
```diff
@@ -28,17 +29,31 @@
 __email__ = 'colour-developers@colour-science.org'
 __status__ = 'Production'
 
-__all__ = ['Lab_to_DIN99', 'DIN99_to_Lab']
+__all__ = ['Lab_to_DIN99', 'DIN99_to_Lab', 'DIN99_METHODS']
 
+DIN99_METHODS = CaseInsensitiveMapping({
+    'ASTMD2244-07':
```
KelSolaar, line 35: Let's also make a `DIN99` variant with the exact same values for reference with the publication.

The paper also say:

> The refined formulas are named DIN99b, DIN99c, and DIN99d.

So...

colour/models/din99.py
```diff
@@ -28,17 +29,31 @@
...
+
+def Lab_to_DIN99(Lab, method='ASTMD2244-07', k_E=1, k_CH=1):
     """
     Converts from *CIE L\\*a\\*b\\** colourspace to *DIN99* colourspace.
 
     Parameters
     ----------
     Lab : array_like
         *CIE L\\*a\\*b\\** colourspace array.
+    method : string, optional
```
KelSolaar, line 54: This should be reordered accordingly and might be better put as follows:

```
    method : unicode, optional
        **{'DIN99', 'ASTMD2244-07', 'DIN99b', 'DIN99c', 'DIN99d'}**,
        Computation method to choose between the :cite:`ASTMInternational2007` formula and the refined formulas according to *Cui et al. (2002)*.
```

colour/models/din99.py
```diff
@@ -85,37 +100,41 @@ def Lab_to_DIN99(Lab, k_E=1, k_CH=1):
     >>> Lab_to_DIN99(Lab)  # doctest: +ELLIPSIS
     array([ 53.2282198...,  28.4163465...,   3.8983955...])
     """
+    p = DIN99_METHODS[str(method)]
```
KelSolaar, line 103: We should validate the method: `p = DIN99_METHODS[validate_method(method, DIN99_METHODS)]`

Maybe `p` should actually be `c`, standing for coefficient and maybe we expand them too, instead of accessing them:

```python
c_1, c_2, c_3, c_4, c_5, c_6, c_7, c_8 = DIN99_METHODS[validate_method(method, DIN99_METHODS)]
```

colour/models/din99.py
```diff
@@ -85,37 +100,41 @@ def Lab_to_DIN99(Lab, k_E=1, k_CH=1):
...
 
     L, a, b = tsplit(to_domain_100(Lab))
 
-    cos_16 = np.cos(np.radians(16))
-    sin_16 = np.sin(np.radians(16))
+    cos = np.cos(np.radians(p[2]))
+    sin = np.sin(np.radians(p[2]))
```
KelSolaar, line 108: `cos` and `sin` feel dangerous here, should we add a `_c` for coefficient suffix? e.g. `cos_c`

colour/models/din99.py
```diff
@@ -85,37 +100,41 @@ def Lab_to_DIN99(Lab, k_E=1, k_CH=1):
...
-    L_99 = 105.509 * (np.log(1 + 0.0158 * L)) * k_E
+    L_99 = p[0] * (np.log(1 + p[1] * L)) * k_E
 
     Lab_99 = tstack([L_99, a_99, b_99])
 
     return from_range_100(Lab_99)
 
 
-def DIN99_to_Lab(Lab_99, k_E=1, k_CH=1):
+def DIN99_to_Lab(Lab_99, method='ASTMD2244-07', k_E=1, k_CH=1):
```
KelSolaar, line 127: All the above comments are valid for the inverse model.

colour/models/din99.py
```diff
@@ -19,7 +19,8 @@
 import numpy as np
```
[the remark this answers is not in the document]
KelSolaar: We might also want to update the module header and the docs:

```
DIN99 Colourspace and DIN99b, DIN99c, DIN99d Refined Formulas
=============================================================

Defines the *DIN99* colourspace and *DIN99b*, *DIN99c*, *DIN99d* refined formulas transformations:
```

Docs: https://github.com/colour-science/colour/blob/259bcae104692293d2a4b0781efc9af6068850c3/docs/colour.models.rst#din99-colourspace

KelSolaar (changes requested):
Thanks @SGeetansh! Some minor things left to address but otherwise LGTM!

Issue #499, opened 2019-09-08
Implement support for the "JzAzBz" based colour appearance model.


KelSolaar:
References
---
- https://www.ingentaconnect.com/content/10.2352/ISSN.2169-2629.2018.26.96

pathakcodes:
@KelSolaar  I am here through GSoC page and interested in implementing new CAM.

I have studied colour.appearance module and can't access the paper as its paid.
Help me get access to it. I am interested in working on this project.

KelSolaar:
Hi Shivam,

Welcome :)

I got back to you via email, see you there!

Cheers,

Thomas

KelSolaar:
Closing in favour of *ZCAM* that was implemented.

PR #850, opened 2021-08-23
PR: Implement support for "Safdar (2018)" colour appearance model.

villirion:
Safdar (2018) is a Jzazbz model who fixed [#499](https://github.com/colour-science/colour/issues/499)


KelSolaar:
Given that we have implemented *ZCAM* and this is its WIP variant, I will close this one!

Issue #654, opened 2020-11-14
Implement support for "OpenColorIO" processor.


KelSolaar:
Colour could use a definition like that to interact easily with OpenColorIO:

```python
def ocio_colourspace_process(a, config, source, target, use_cpu=True):
    a = colour.utilities.as_float_array(a, dtype=np.float32)
    height, width, channels = a.shape

    processor = config.getProcessor(source, target)
    processor = (
        processor.getDefaultCPUProcessor()
        if use_cpu else
        processor.getDefaultGPUProcessor()
    )

    image_desc = ocio.PackedImageDesc(a,
        width, height, channels)

    processor.apply(image_desc)

    return image_desc.getData().reshape([height, width, channels])
```

The issue implementing support for that is mostly logistical as it already works but it:

- Might be relevant to wait for movement on AcademySoftwareFoundation/OpenColorIO#1206
- Requires OpenColorIO artifacts on Github Actions
- Requires a minimal OpenColorIO config
- Requires unit tests
- Requires CI update

zachlewis:
I'd also love to see a variant that takes display and view arguments instead of a target color space:

```
def ocio_display_view_process(a, config, source, display=None, view=None, use_cpu=True):
    a = colour.utilities.as_float_array(a, dtype=np.float32)
    height, width, channels = a.shape

    processor = config.getProcessor(source, display=display, view=view)
    ...


```

PR #869, opened 2021-09-13
PR: Implement support for "OpenColorIO" processor.

KelSolaar:
References #654.

The API is as follows:

```python
>>> config = os.path.join(os.path.dirname(__file__), 'tests', 'resources',  'config-aces-reference.ocio.yaml')
>>> a = full([4, 2, 3], 0.18)
>>> process_image_OpenColorIO(a, 'ACES - ACES2065-1', 'ACES - ACEScct', config=config)
array([[[ 0.4135878...,  0.4135878...,  0.4135878...],
        [ 0.4135878...,  0.4135878...,  0.4135878...]],

       [[ 0.4135878...,  0.4135878...,  0.4135878...],
        [ 0.4135878...,  0.4135878...,  0.4135878...]],

       [[ 0.4135878...,  0.4135878...,  0.4135878...],
        [ 0.4135878...,  0.4135878...,  0.4135878...]],

       [[ 0.4135878...,  0.4135878...,  0.4135878...],
        [ 0.4135878...,  0.4135878...,  0.4135878...]]], dtype=float32)
>>> process_image_OpenColorIO(a, 'ACES - ACES2065-1', 'Display - sRGB',  'Output - SDR Video - ACES 1.0', ocio.TRANSFORM_DIR_FORWARD, config=config)
array([[[ 0.3559523...,  0.3559525...,  0.3559525...],
        [ 0.3559523...,  0.3559525...,  0.3559525...]],

       [[ 0.3559523...,  0.3559525...,  0.3559525...],
        [ 0.3559523...,  0.3559525...,  0.3559525...]],

       [[ 0.3559523...,  0.3559525...,  0.3559525...],
        [ 0.3559523...,  0.3559525...,  0.3559525...]],

       [[ 0.3559523...,  0.3559525...,  0.3559525...],
        [ 0.3559523...,  0.3559525...,  0.3559525...]]], dtype=float32)
```

it is also possible to use the *OCIO* config from the environment.

### DIFF
--- a/colour/io/ocio.py
+++ b/colour/io/ocio.py
@@ -39,7 +39,9 @@ def process_image_OpenColorIO(a, *args, **kwargs):
         See https://opencolorio.readthedocs.io/en/latest/api/config.html for
         more information.
     config : unicode, optional
-        *OpenColorIO* config to use for processing.
+        *OpenColorIO* config to use for processing. If not defined, the
+        *OpenColorIO* set defined by the ``$OCIO`` environment variable is
+        used.
 
     Returns
     -------
--- a/colour/io/ocio.py
+++ b/colour/io/ocio.py
@@ -40,8 +40,6 @@ def process_image_OpenColorIO(a, *args, **kwargs):
         more information.
     config : unicode, optional
         *OpenColorIO* config to use for processing.
-    use_cpu : bool, optional
-        Whether to use the *CPU* or *GPU* processor.
 
     Returns
     -------
@@ -89,8 +87,6 @@ def process_image_OpenColorIO(a, *args, **kwargs):
 
     import PyOpenColorIO as ocio
 
-    use_cpu = kwargs.get('use_cpu', True)
-
     config = kwargs.get('config')
     config = (ocio.Config.CreateFromEnv()
               if config is None else ocio.Config.CreateFromFile(config))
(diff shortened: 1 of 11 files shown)


colour/io/ocio.py
```diff
@@ -0,0 +1,110 @@
...
+    config = (ocio.Config.CreateFromEnv()
+              if config is None else ocio.Config.CreateFromFile(config))
+
+    a = np.atleast_3d(as_float_array(a, dtype=np.float32))
+    height, width, channels = a.shape
+
+    processor = config.getProcessor(*args)
+
+    processor = (processor.getDefaultCPUProcessor()
+                 if use_cpu else processor.getDefaultGPUProcessor())
```
zachlewis, line 104: I don't think you can use a GPU Processor like you can a CPU Processor; i.e., there's no "apply" method....

KelSolaar: I actually haven't tried nor looked at it! I will dig through the OCIO codebase.

KelSolaar: So looks like it is more involved indeed: https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/0021372b4dba3444e21a36d846b327efad054f8d/tests/python/GpuShaderDescTest.py#L127

remia: And on top of that you need to create an OpenGL context, manually upload any texture data needed and execute your draw call...!

KelSolaar: Yeah, I saw that, will try to update that tonight, be great to consider making that easier, there is probably no good reasons why it cannot be abstracted for a user, c.f. `import cupy as np`.

KelSolaar: I went path-of-least-resistance and removed the non-working GPU path, we can figure that out in a later PR.

KelSolaar: So, confirming that we will not have GPU processing anytime soon: https://github.com/AcademySoftwareFoundation/OpenColorIO/pull/1493/commits/4a8dac4affef4fdea3c2c5b4d67abde7c803c645

KelSolaar:
Merging this one!

Issue #792, opened 2021-03-12
The *.spimtx reader in the "feature/clf" branch reads offsets incorrectly.


nick-shaw:
https://github.com/colour-science/colour/blob/6638055709a87162dd1a738bc8ec2101d99d75ad/colour/io/luts/sony_spimtx.py#L36

Looking at the [.spimtx parser in OCIO](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/9b23e9623792d8cc6e6c1dfd5394335ee148bcf3/src/OpenColorIO/fileformats/FileFormatSpiMtx.cpp#L134), it would appear that where a .spimtx file includes offsets, these should be treated as 16-bit integer scaled, not normalised float. Likewise the .spimtx writer should do the inverse. When writing the .spimtx parser, I had assumed incorrectly that the format used float offsets, the same as the matrix operator in CLF.

KelSolaar:
This one was addressed a long time ago!

PR #890, opened 2021-10-29
PR: Implement support for "SPImtx" LUT input/output.

KelSolaar:
This PR implements support for *SPImtx* input and output via the `colour.io.read_LUT_SonySPImtx` and `colour.io.write_LUT_SonySPImtx` definition.

It is based on the work of @nick-shaw in https://github.com/colour-science/colour/commit/1decfe0ca4ad3b2ce6b20828f31ad1206f54c298.

I would like to discuss about the `colour.io.Matrix` class name, ideally it should be prefixed to something like this `LUTSequenceOperatorMatrix`. I considered renaming `AbstractLUTSequenceOperator` to `AbstractLUTOperator` which would then yield `LUTOperatorMatrix`.


colour/io/luts/sony_spimtx.py
```diff
@@ -0,0 +1,101 @@
...
+    --------
+    >>> matrix = np.array([[ 1.45143932, -0.23651075, -0.21492857],
+    ...                    [-0.07655377,  1.1762297 , -0.09967593],
+    ...                    [ 0.00831615, -0.00603245,  0.9977163 ]])
+    >>> M = Matrix(matrix)
+    >>> write_LUT_SonySPI1D(M, 'My_LUT.spimtx')  # doctest: +SKIP
+    """
+
+    if matrix.matrix.shape == (3, 4):
+        matrix = matrix.matrix
```
zachlewis, line 95: Per #792, we should scale the offset by 65535. (The OCIO spimtx reader doesn't seem to mind if the actual values are cast as floats)

zachlewis (changes requested):
Looks awesome. 

For spimtx io, we should scale by 65535; otherwise, this is perfect

colour/io/luts/sony_spimtx.py
KelSolaar: Thx, I checked the OCIO parser after opening the PR and it raised some more questions, so I stashed for today and opened the PR nonetheless.

Here is some context:

- For SPImtx, the offsets are currently in the last column.
- The new `Matrix` class has separate matrix + offset, i.e. `Matrix.matrix` and `Matrix.offset`.
  - If the matrix is 3x4, the last column acts on a pseudo-channel in the current implementation, i.e. a new white channel concatenated.
    - **Issue**: We never handle alpha in Colour, but for that particular case, we add an alpha channel if the `Matrix.matrix` is 3x4 but we might also probably handle the case where RGB might actually be RGBA, i.e. (..., 4).
    - OCIO applies the 4x4 matrix on RGBA also.
- **Issue**: With that in mind, we simply cannot read the SPImtx matrix directly and shove it in `Matrix.matrix` because then the offset will apply on RGB(**A**) thus we should extract the offset, i.e. last column and put it in  `Matrix.offset`.

Thoughts?

KelSolaar: Relevant bit for discussion: https://opencolorio.readthedocs.io/en/latest/api/transforms.html#matrixtransform

image

zachlewis: Ah yeah -- I think spimtx only encodes twelve values, and none of them are alpha. And when I peeked at the CTL parser, I think I saw something about 5x5 matrices and tried to put it out of my mind. 

Perhaps we should extract the offset and apply as a separate operation on top of the 4x4 or 3x3, input shape pending.

If it's helpful, here's what I'm using to dump PyOpenColorIO.MatrixTransforms to spimtx files:

```
def write_spimtx(xform: ocio.MatrixTransform, path: Union[str, Path], decimals: int = 9):
    matrix, offset = map(as_float_array, (xform.getMatrix(), xform.getOffset()))
    offset *= 65535
    array = np.hstack([np.reshape(matrix, (4, 4))[0:3, 0:3], np.array([offset[0:3]]).T])
    np.savetxt(path, array, fmt='%.{0}f'.format(decimals).encode('utf-8'))
```

KelSolaar: I will update our code accordingly, what about the `Matrix` class name itself? Any strong opinions?

zachlewis: No strong opinions... maybe `MatrixOffset`? (Which we could perhaps one day contrast with `OffsetMatrix`, a la Daniele's Academy Linear Grade proposal :) )

KelSolaar: Ah good point! I was referring to the OP where suggested to adopt a longer `LUTOperatorMatrix` name for easier discovery but then `LUTOperatorMatrixOffset`? :D

zachlewis: Ha -- duh, the OP. Of course. 

`LUTOperatorMatrix` with an argument for applying the offset first sounds great. And I guess for writing spimtxs when the offset is applied first, we could either throw or... offer to write two spimtxs or something :)

KelSolaar: Just when we thought it would be easy...

image

KelSolaar: If we have the argument to the `LUTOperatorMatrix.apply` method it simplifies things a bit, I will put it there for now as it also separates concerns a bit.

zachlewis: That sounds perfect. 👍 👍 👍

KelSolaar: Think we tick the boxes with https://github.com/colour-science/colour/pull/890/commits/d02ae76273ed34dfd09d50129dcda8bfc09f0422

KelSolaar:
I implement a few changes as per-above discussion with @zachlewis:

- The renamed `colour.io.LUTOperatorMatrix` should now handle 3x3 and 4x4 matrices and offset.
- Internally, everything is promoted to (4, 4) and (4, ) shapes as it simplifies the code.
- The SPImtx IO code has been updated accordingly.

I will probably add some tests for the RGBA case as I realise I did not put any and I will probably expose `colour.io.LUTOperatorMatrix` in the upper `colour` namespace, i.e. `colour.LUTOperatorMatrix`.

zachlewis (approved):
I love it. Ship it!

Issue #812, opened 2021-04-30
Implement support for "ZCAM" colour appearance model.


KelSolaar:
References
---
- https://www.osapublishing.org/oe/fulltext.cfm?uri=oe-29-4-6036&id=447640
- https://github.com/ksmet1977/luxpy/issues/20

PR #894, opened 2021-11-06
PR: Implement support for "ZCAM" colour appearance model.

KelSolaar:
This PR implements support the *ZCAM* colour appearance model by *Safdar, Hardeberg and Luo (2021)*.

A few things of interest and as written in the docstrings:

- *Safdar, Hardeberg and Luo (2021)* does not specify how the chromatic adaptation to *CIE Standard Illuminant D65* in *Step 0* should be performed. A one-step *Von Kries* chromatic adaptation transform is not or transitive when a degree of adptation is involved. *Safdar, Hardeberg and Luo (2018)* uses *Zhai and Luo (2018)* two-steps chromatic adaptation transform, thus it seems sensible to adopt this transform for the *ZCAM* colour appearance model until more information is available. It is worth noting that a one-step *Von Kries* chromatic adaptation transform with support for degree of adaptation produces values closer to the supplemental document compared to the *Zhai and Luo (2018)* two-steps chromatic adaptation transform but then the *ZCAM* colour appearance model does not round-trip properly.
- *Step 4* of the inverse model uses a rounded exponent of 1.3514 preventing the model to round-trip properly. Given that this implementation takes some liberties with respect to the chromatic  transform to use, it was deemed appropriate to use an exponent value, i.e. 50 / 37, that enables the *ZCAM* colour appearance model to round-trip.
- The values in the third column of the supplemental document are likely incorrect:
    - Hue quadrature `H_z` is significantly different for this test, i.e. 47.748252 vs 43.8258.
    - `F_L` as reported in the supplemental document has the same value as for `L_a` = 264 instead of 150. 

References #812.


KelSolaar:
Going to merge this one, it matches quite well the supplemental paper except for the above points that we will revisit when more information is available. I will send an email to some of the authors.

PR #933, opened 2022-01-30
PR: Add __eq__, __ne__, and __hash__ methods for "colour.io.tm2714.Header_IESTM2714" class.

JGoldstone:
Fairly straightforward, but: please consider whether it suffices to just compare the complete set of properties or whether the mapping attributes need also be compared.

### DIFF
--- a/colour/io/tests/test_tm2714.py
+++ b/colour/io/tests/test_tm2714.py
@@ -3,13 +3,13 @@ Defines the unit tests for the :mod:`colour.io.tm2714` module.
 """
 
 from __future__ import annotations
-from copy import deepcopy
 
 import numpy as np
 import os
 import shutil
 import unittest
 import tempfile
+from copy import deepcopy
 
 from colour.colorimetry import SpectralDistribution
 from colour.hints import Dict, List, Optional, Tuple, Union, cast
@@ -173,7 +173,7 @@ class TestIES_TM2714_Header(unittest.TestCase):
 
     def test__eq__(self):
         """
-        Tests :meth:`colour.io.tm2714.Header_IESTM2714.__eq__ method.
+        Tests :meth:`colour.io.tm2714.Header_IESTM2714.__eq__` method.
         """
 
         h0 = Header_IESTM2714(manufacturer='a', catalog_number='b', description='c',
@@ -187,7 +187,7 @@ class TestIES_TM2714_Header(unittest.TestCase):
 
     def test__ne__(self):
         """
-        Tests :meth:`colour.io.tm2714.Header_IESTM2714.__ne__ method.
+        Tests :meth:`colour.io.tm2714.Header_IESTM2714.__ne__` method.
         """
 
         h0 = Header_IESTM2714(manufacturer='a', catalog_number='b', description='c',
@@ -204,7 +204,7 @@ class TestIES_TM2714_Header(unittest.TestCase):
 
     def test__hash__(self):
         """
-        Tests :meth:`colour.io.tm2714.Header_IESTM2714.__hash__ method.
+        Tests :meth:`colour.io.tm2714.Header_IESTM2714.__hash__` method.
         """
 
         h0 = Header_IESTM2714(manufacturer='a', catalog_number='b', description='c',
--- a/colour/io/tests/test_tm2714.py
+++ b/colour/io/tests/test_tm2714.py
@@ -3,6 +3,7 @@ Defines the unit tests for the :mod:`colour.io.tm2714` module.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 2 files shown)


colour/io/tests/test_tm2714.py
```diff
@@ -3,6 +3,7 @@
 """
 
 from __future__ import annotations
+from copy import deepcopy
```
KelSolaar, line 6: We usually put a `from ... import` just after the regular `import`, here after `import os`

colour/io/tests/test_tm2714.py
```diff
@@ -170,6 +171,50 @@ def test_required_attributes(self):
         for attribute in required_attributes:
             self.assertIn(attribute, dir(Header_IESTM2714))
 
+    def test__eq__(self):
+        """
+        Tests :meth:`colour.io.tm2714.Header_IESTM2714.__eq__ method.
```
KelSolaar, line 176: The backtick is missing at the end here ``:meth:`colour.io.tm2714.Header_IESTM2714.__eq__`` and on all the subsequent docstrings.

colour/io/tm2714.py
```diff
@@ -601,6 +601,34 @@ def comments(self, value: Optional[str]):
 
         self._comments = value
 
+    def __eq__(self, other):
```
KelSolaar, line 604: We will need the typing annotations and docstrings here please, a good example: https://github.com/colour-science/colour/blob/develop/colour/continuous/signal.py#L835

colour/io/tm2714.py
```diff
@@ -601,6 +601,34 @@ def comments(self, value: Optional[str]):
...
+                self._report_number == other.report_number,
+                self._report_date == other.report_date,
+                self._document_creation_date == other.document_creation_date,
+                self._comments == other.comments
+            ]
+            return all(comps)
+        return False
+
+    def __ne__(self, other):
+        return not self == other
```
KelSolaar, line 623: I had to double-check the priority of `==` compared to `not`, for clarity, I tend to parentheses:

```python
    def __ne__(self, other):
        return not (self == other)
```

KelSolaar (changes requested):
Thanks @JGoldstone, a few minor things but it looks good otherwise!

JGoldstone:
Addressed, thank you.

Is it the case that all I need to do is to push to the same remote from which the PR was made, and then you see the updates automatically? Or need I do something else?

KelSolaar:
Yes correct!

JGoldstone:
Well, that’s done, then. It took another iteration to get the return type hint from __hash__ right. And yet…the macOS CI build is thinking that I got it wrong and that __hash__ should return a bool, or that’s how I interpret this:

https://github.com/JGoldstone/colour/runs/5006350350?check_suite_focus=true <https://github.com/JGoldstone/colour/runs/5006350350?check_suite_focus=true>

Would appreciate if you could take a look.

KelSolaar:
So the method as written returns a `bool`:

```python
    def __hash__(self) -> bool:
        return hash((self._manufacturer, self._catalog_number, self._description,
                      self._document_creator, self._unique_identifier,
                      self._measurement_equipment, self._laboratory,
                     self._report_number, self._report_date,
                     self._document_creation_date, self._comments))
```

But the output of the `hash` definition is an `int`, so `__hash__ ` should be typed with `int` and not `bool`.

Not that I made aliases for various things, for example we have `Boolean` for `bool` or `Integer` for `int`. The rational for that is originally, when I started to add annotations, those type hints were defined roughly as follows:

```
Integer = Union[int, np.int8, np.int16, , np.int32, , np.int64, ...]
Boolean = Union[bool, np.bool_]
```

Now, because of some difficulty in making Mypy/Numpy think that the unions above should be valid in all cases, I have reduced them to:

```
Integer = int
Boolean = bool
```

I however kept the aliasing because it is useful to have that indirection. With that in mind, it would be great if you could use them instead of `int` and `bool`.

KelSolaar:
I will merge this one and add the missing docstrings in a subsequent commit! Thanks @JGoldstone!

JGoldstone:
Thanks for all the extra cleanup on docstrings and on tighter style

And with this, I even managed to take advantage of your having merged the PR () without having to delete my fork and re-fork, as has been my sorry wont in the past:
http://vakila.github.io/blog/my-pr-was-merged-now-what/ <http://vakila.github.io/blog/my-pr-was-merged-now-what/>…so it’s feeling like a pretty decent morning.

In re: having to fall back to using the simple rather than Union types, will that get easier in Python 3.10 with its inclusion of PEP 604 <https://www.python.org/dev/peps/pep-0604/>?

Issue #982, opened 2022-05-13
[FEATURE]: Aliases for ffmpeg tone curves and primaries


fxthomas:
Ffmpeg provides a large set of primaries and transfer functions (OETFs) for use as metadata in video formats, with the `-color_trc` and `-color_primaries` parameters (or equivalent `AVFrame` members).

Allowed values for these options are stated to come out of [ISO/IEC 23001-8:2013](https://www.iso.org/standard/62088.html) in [libavutil/pixfmt.h](https://github.com/FFmpeg/FFmpeg/blob/master/libavutil/pixfmt.h#L471), and reference existing standards and recommendations that are for the most part already implemented in colour-science.

Is there any interest in adding aliases to let people reference these more easily?

KelSolaar:
Hi @fxthomas,

I don't have a personal use case but it could be useful. We would need to source the standard though. Here are a few pages: https://cdn.standards.iteh.ai/samples/69661/48c09dfd01a840e2b7c99c70dc5338b1/ISO-IEC-23001-8-2016.pdf

fxthomas:
I'm willing to create PR for that, how should I organize this within the colour package? Would something like the following work?

The standard explicitly define these:

* `colour.models.rgb.datasets.iso_iec_23001_8_2013.Iso23001ColourPrimaries`: An integer enum containing the keys defined in §7.1
* `colour.models.rgb.datasets.iso_iec_23001_8_2013.PRIMARIES_ISO_IEC_23001_8_2013`: a `dict` mapping `Iso23001ColourPrimaries` values to the actual primaries
* `colour.models.rgb.datasets.iso_iec_23001_8_2013.CCS_WHITEPOINTS_ISO_IEC_23001_8_2013`: a `dict` mapping `Iso23001ColourPrimaries` values to the white point chromaticity coordinates
* `colour.models.rgb.transfer_functions.iso_iec_23001_8_2013.Iso23001TransferCharacteristics`: An integer enum containing keys defined in §7.2
* `colour.models.rgb.transfer_functions.iso_iec_23001_8_2013.OETFS_ISO_IEC_23001_8_2013`: a `dict` mapping `Iso23001TransferCharacteristics` values to the actual OETFs

The following can be derived from the standard and should IMO be also added for consistency with the other modules:

[...]

PR #987, opened 2022-05-18
PR: Add aliases for ITU-T H.273 video color metadata

fxthomas:
# Summary

This implements what's discussed in #982.

I tried to stick as close to the original ISO standard as possible, but there are a few caveats:

- Some sets of primaries/whitepionts as well as transfer functions were not already present in colour ; I implemented them here based on the ISO standard in this module and added the necessary references.
- The ISO standard is not very clear on the values of alpha, beta and gamma for ITU-R BT.1361, so I chose instead to implement the OETF based on that standard.
- I'm not very certain of how I should implement EOTFs/OETFs when the range is not [0-1] (e.g. BT.1361, which has an extended range). Is there any consensus on that? How is it usually documented?
- I linked the RGB <-> XYZ matrices from the original module in colour even though they are not defined explicitly in the ISO standard, just because it makes things more consistent with how the original modules in `colour.models.rgb.datasets` work. Let me know if this is not needed.

# Preflight

**Personal todo**
- [x] Verify that ISO 23001-8:2013 aligns with ISO 23091-2
- [x] Verify that ISO 23001-8:2013 aligns with ITU-T H.273
- [x] Verify that ISO [14496-10:2020](https://standards.iso.org/ittf/PubliclyAvailableStandards/index.html) aligns with the rest
- [x] Finish updating the doctests/examples
- [x] Finish adding typing hints

**Code Style and Quality**

- [ ] New transformations have been added to the *Automatic Colour Conversion Graph*.

**Documentation**

### DIFF
--- a/colour/models/rgb/datasets/ebu_3213_e.py
+++ b/colour/models/rgb/datasets/ebu_3213_e.py
@@ -1,0 +1,64 @@
+"""
+EBU Tech. 3213-E Primaries and White Point
+==========================================
+
+Defines primaries and white point chromaticity coordinates from EBU Tech.
+3213-E (:cite:`EuropeanBroadcastingUnion3213E`).
+
+References
+----------
+-   :cite:`EuropeanBroadcastingUnion3213E` : European Broadcasting Union
+    Tech. 3213-E (1975), Standard for Chromaticity Tolerances For Studio
+    Monitors.
+    https://tech.ebu.ch/docs/tech/tech3213.pdf
+"""
+
+from __future__ import annotations
+
+import numpy as np
+
+from colour.hints import NDArray
+from colour.models.rgb import normalised_primary_matrix
+
+__author__ = "Colour Developers"
+__copyright__ = "Copyright 2013 Colour Developers"
+__license__ = "New BSD License - https://opensource.org/licenses/BSD-3-Clause"
+__maintainer__ = "Colour Developers"
+__email__ = "colour-developers@colour-science.org"
+__status__ = "Production"
+
+__all__ = [
+    "PRIMARIES_EBU_3213_E",
+    "WHITEPOINT_NAME_EBU_3213_E",
+    "CCS_WHITEPOINT_EBU_3213_E",
+    "MATRIX_EBU_3213_E_RGB_TO_XYZ",
+    "MATRIX_XYZ_TO_EBU_3213_E_RGB",
+]
+
+PRIMARIES_EBU_3213_E: NDArray = np.array(
+    [
+        [0.64, 0.33],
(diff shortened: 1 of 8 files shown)


KelSolaar:
> Some sets of primaries/whitepionts as well as transfer functions were not already present in colour ; I implemented them here based on the ISO standard in this module and added the necessary references.

I think we will probably want every single one in its own module for consistency, will make surfacing them into the API easier too.

> I'm not very certain of how I should implement EOTFs/OETFs when the range is not [0-1] (e.g. BT.1361, which has an extended range). Is there any consensus on that? How is it usually documented?

For some of them we have `in_normalised_code_value` / `out_normalised_code_value` kwargs that perform full to legal / legal to full conversion.

fxthomas:
> I think we will probably want every single one in its own module for consistency, will make surfacing them into the API easier too.

Okay, will do that.

> For some of them we have in_normalised_code_value / out_normalised_code_value kwargs that perform full to legal / legal to full conversion.

I think i will need to look into it a bit more to be sure of what I'm doing, because technically these are beyond-full-range if I understand what their extended range means. This is similar to transfer characteristics that accept a floats in a [-1, 1] range, technically they're not "full-range", are they?

I'll work on that PR as soon as I get some time, whose availability was pretty scarce the last couple of weeks ;)

fxthomas:
Okay, should be good for review, I think.

* As discussed with the other stories, the main definitions reference ITU-T H.273, since it's the most accessible standard for these metadata.

* The `color.models.rgb.video` module contains only mappings and all the other TFs and primaries are defined in new modules in `color.models.rgb.datasets`/`transfer_functions` when needed.

* I've tried to be as precise as possible with the references ; for those I don't have direct access to (IEC 61966 and ST 428) I added a note in the docstring specifying that I used the ITU-T H.273 definition, hopefully somebody will later come along and check that what I did is indeed the same as what's defined in the referenced standard.

KelSolaar:
Hi @fxthomas,

Thanks for the great work! I think I will probably merge it in a staging branch as it requires a polishing pass! Something that would be really helpful would be if you could implement the inverse of the new CCTF you added please?

Cheers,

Thomas

fxthomas:
@KelSolaar Sure! I added the inverses for all the new OETFs. Also when checking ITU-R BT.1361 I found a typo in the earlier commits (I had swapped the two lines in the `np.where`), which I also corrected.

Let me know if you need anything else.

KelSolaar:
This is awesome, thanks a lot @fxthomas, I will review that over the weekend!

KelSolaar:
@fxthomas : I will merge in a staging branch and varnish things a bit, thanks a lot!

PR #994, opened 2022-06-02
Feature/tjd test improvements

tjdcs:
# Summary

This is a first contribution PR mainly to practice the branching, PR, testing, and general development tools of colour-science. It contains some minor changes to the testing infrastructure. It also improves the developer experience using vscode. 

* The tasks.py uses `pytest` instead of `pytest` (deprecated). 
* `matplotlib` is configured into headless mode with a pytest fixture instead of an environment variable (improves usability of vscode testing tools).
* The testing task lists the 5 longest duration tests (good to be mindful)
* One extremely long running test has been reduced from 4 minutes to a few seconds (MBP2020 4-core i5)
* Test run on multiple workers by default with pytest-xdist

# Preflight

**Code Style and Quality**

**Documentation**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,5 @@ docs/_static/Tutorial_*.png
 docs/generated
 poetry.lock
 references
+
+.vscode/
--- a/colour/plotting/conftest.py
+++ b/colour/plotting/conftest.py
@@ -1,0 +1,29 @@
+"""
+Plotting - Tests
+================
+Configures pytest to use the AGG headless backend. This allows the plotting
+unittests to run without creating windows in IDEs such as VSCode.
+"""
+import pytest
+import matplotlib
+
+__author__ = "Colour Developers"
+__copyright__ = "Copyright 2013 Colour Developers"
+__license__ = "New BSD License - https://opensource.org/licenses/BSD-3-Clause"
+__maintainer__ = "Colour Developers"
+__email__ = "colour-developers@colour-science.org"
+__status__ = "Production"
+
+
+@pytest.fixture(autouse=True, scope="session")
+def mplHeadlessBackend() -> None:
+    """
+    Configure matplotlib for headless testing.
+
+    pytest Fixture automatically applied to any tests in this package or any
+    subpackages at the beginning of the pytest session.
+    """
+    curBackend = matplotlib.get_backend()
+    matplotlib.use("AGG")
+    yield
+    matplotlib.use(curBackend)
--- a/colour/plotting/conftest.py
+++ b/colour/plotting/conftest.py
@@ -4,6 +4,7 @@ Plotting - Tests
 Configures pytest to use the AGG headless backend. This allows the plotting
 unittests to run without creating windows in IDEs such as VSCode.
 """
+from typing import Generator
 import pytest
 import matplotlib
 
@@ -16,7 +17,7 @@ __status__ = "Production"
 
 
 @pytest.fixture(autouse=True, scope="session")
-def mplHeadlessBackend() -> None:
+def mplHeadlessBackend() -> Generator[None, None, None]:
     """
     Configure matplotlib for headless testing.
 
--- a/colour/volume/tests/test_spectrum.py
+++ b/colour/volume/tests/test_spectrum.py
@@ -3,7 +3,7 @@
 
 import numpy as np
 import unittest
-from itertools import permutations
+from itertools import product
 
 from colour.colorimetry import (
     MSDS_CMFS,
@@ -243,9 +243,9 @@ class TestIsWithinVisibleSpectrum(unittest.TestCase):
         """
 
         cases = [-1.0, 0.0, 1.0, -np.inf, np.inf, np.nan]
-        cases = set(permutations(cases * 3, r=3))
-        for case in cases:
-            is_within_visible_spectrum(case)
+        cases = set(product(cases, repeat=3)) - {(0, 0, 0)}
+        isVisible = is_within_visible_spectrum(list(cases))
+        assert all(~isVisible)
 
 
 if __name__ == "__main__":
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -77,6 +77,7 @@ sphinx = { version = ">= 4, < 5", optional = true }  # Development dependency.
 sphinxcontrib-bibtex = { version = "*", optional = true }  # Development dependency.
 toml = { version = "*", optional = true }  # Development dependency.
 twine = { version = "*", optional = true }  # Development dependency.
+pytest-xdist = { version = "^2.5.0", optional = true } #Development dependency.
 
 [tool.poetry.dev-dependencies]
 biblib-simple = "*"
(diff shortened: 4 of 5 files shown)


tjdcs:
The changes i made to testing `is_within_visible_spectrum` possibly revealed a bug where `(1,1,1)` is not considered within the spectrum for the default arguments if illuminant E and 2-deg obs. I believe that it probably should be though, since a perfect diffuse reflector of illuminant E should produce X = Y = Z = 1.

Also it seems like other values, such as X = Y = Z = 2 should be considered in the spectrum because they are realizable values, but just brighter than the white point. In practice, this can happen when you are normalizing XYZ based on a PRD but your dataset contains values representing non-diffuse reflections which can be brighter than the whitepoint. 

I believe I read that @KelSolaar has a paper related to this code so perhaps they can ponder on this. In any case the changes to the test just make it much much faster and preserve what I believe was the real objective of this test case.

KelSolaar:
@tjdcs: What you see is related to precision issues from the coordinates we use and the tessellation `QHull` produces under the hood:

image

The tolerance is currently set to 0 but I reckon we could set it to the minimum value so that 1 is accepted. Note that we are purposely not handling *HDR* values for that particular definition and the expectation is that they are normalised (which we should indicate in the docstrings).

colour/plotting/conftest.py
```diff
@@ -0,0 +1,30 @@
...
+__author__ = "Colour Developers"
+__copyright__ = "Copyright 2013 Colour Developers"
+__license__ = "New BSD License - https://opensource.org/licenses/BSD-3-Clause"
+__maintainer__ = "Colour Developers"
+__email__ = "colour-developers@colour-science.org"
+__status__ = "Production"
+
+
+@pytest.fixture(autouse=True, scope="session")
+def mplHeadlessBackend() -> Generator[None, None, None]:
```
KelSolaar, line 20:
```suggestion
def mpl_headless_backend() -> Generator[None, None, None]:
```

colour/plotting/conftest.py
```diff
@@ -0,0 +1,30 @@
...
+import matplotlib
+
+__author__ = "Colour Developers"
+__copyright__ = "Copyright 2013 Colour Developers"
+__license__ = "New BSD License - https://opensource.org/licenses/BSD-3-Clause"
+__maintainer__ = "Colour Developers"
+__email__ = "colour-developers@colour-science.org"
+__status__ = "Production"
+
+
```
KelSolaar, line 18:
```suggestion
__all__ = [
    "mpl_headless_backend",
]

```

colour/plotting/conftest.py
```diff
@@ -0,0 +1,30 @@
+"""
+Plotting - Tests
+================
+Configures pytest to use the AGG headless backend. This allows the plotting
+unittests to run without creating windows in IDEs such as VSCode.
+"""
+from typing import Generator
+import pytest
+import matplotlib
```
KelSolaar, line 9:
```suggestion

import matplotlib
import pytest
from typing import Generator
```

.gitignore
```diff
@@ -17,3 +17,5 @@ docs/_static/Tutorial_*.png
 docs/generated
 poetry.lock
 references
+
+.vscode/
```
KelSolaar, line 21: This one should be after `.mypy_cache`, we probably don't need the last slash too.

colour/plotting/conftest.py
```diff
@@ -0,0 +1,30 @@
+"""
+Plotting - Tests
+================
+Configures pytest to use the AGG headless backend. This allows the plotting
+unittests to run without creating windows in IDEs such as VSCode.
```
KelSolaar, line 5:
```suggestion
Plotting - Pytest Fixtures
==========================

Configures *pytest* to use *Matplotlib* *AGG* headless backend. This allows the plotting
unittests to run without creating windows in IDEs such as *VSCode*.
```

colour/plotting/conftest.py
```diff
@@ -0,0 +1,30 @@
...
+__status__ = "Production"
+
+
+@pytest.fixture(autouse=True, scope="session")
+def mplHeadlessBackend() -> Generator[None, None, None]:
+    """
+    Configure matplotlib for headless testing.
+
+    pytest Fixture automatically applied to any tests in this package or any
+    subpackages at the beginning of the pytest session.
```
KelSolaar, line 25:
```suggestion
    Configure *Matplotlib* for headless testing with the *AGG* backend.

    This *pytest* fixture is automatically applied to any tests in this package
    or any sub-packages at the beginning of the *pytest* session.

    Yields
    ------
    Generator
        *Matplotlib* unit tests.
```

colour/plotting/conftest.py
```diff
@@ -0,0 +1,30 @@
...
+    """
+    curBackend = matplotlib.get_backend()
+    matplotlib.use("AGG")
+    yield
+    matplotlib.use(curBackend)
```
KelSolaar, line 30:
```suggestion
    current_backend = matplotlib.get_backend()
    matplotlib.use("AGG")
    yield
    matplotlib.use(current_backend)
```

colour/volume/tests/test_spectrum.py
```diff
@@ -243,9 +243,9 @@ def test_nan_is_within_visible_spectrum(self):
         """
 
         cases = [-1.0, 0.0, 1.0, -np.inf, np.inf, np.nan]
-        cases = set(permutations(cases * 3, r=3))
-        for case in cases:
```
KelSolaar, line 247: One reason for the iteration case by case was that it is easier to figure out when the tested definition breaks. It is slower that being said, will need to think about it as we also have plenty of them.

pyproject.toml
```diff
@@ -93,6 +93,7 @@ pydata-sphinx-theme = "*"
 pydocstyle = "*"
 pytest = "*"
 pytest-cov = "*"
+pytest-xdist = "^2.5.0"
```
KelSolaar, line 96:
```suggestion
pytest-xdist = "*"
```

KelSolaar: We will need it in the two other places (one above, one below).

KelSolaar (changes requested):
Thanks @tjdcs! A few minor suggested edits, the one thing I need to think about is the `nan` thing.

colour/volume/tests/test_spectrum.py
tjdcs: This uses a *huge* amount of time (I think because of setup in the underlying code?) I know the old adage is that machine time is *free-ish* and developer time is very expensive but this seems much to excessive to me. Besides, I like running tests. And hate waiting too long. 

If it fails, then presumably developers are going to enter this function anyway and sort out the failing case very quickly.

pyproject.toml
tjdcs: Corrected in e2d0255c48e33c873a459fbfa1574659d89df3a5

colour/plotting/conftest.py
tjdcs: c8a6ac3aec299da6d9305247ad24129a669c7412

colour/plotting/conftest.py
tjdcs: c8a6ac3aec299da6d9305247ad24129a669c7412

.gitignore
KelSolaar: Fine by me, bonus points for the `.` to hide it! :)

KelSolaar:
Merged this one, thanks a lot! I will need to propagate some of the changes to our other repos.

PR #1036, opened 2022-08-25
PR: Signature update of spectral to XYZ definitions in the context of absolute colorimetry computations.

tjdcs:
# Summary

k is passed by position in several functions, so the order of arguments matters. Bug fix. 

in discrete integration, you also need to multiply by ∆λ 

adds unit tests

# Preflight

**Code Style and Quality**

**Documentation**


colour/colorimetry/tests/test_tristimulus_values.py
```diff
@@ -13,6 +13,7 @@
 
 import numpy as np
 import unittest
+from pytest import approx
```
KelSolaar, line 16: **Major**

At this moment and because *pytest* is not a hard dependency, we should not depend on it and we should use `self.assertAlmostEqual` or `np.testing.assert_array_almost_equal`.

colour/colorimetry/tests/test_tristimulus_values.py
```diff
@@ -1320,6 +1322,91 @@ def test_raise_exception_sd_to_XYZ_ASTME308(self):
         )
 
 
+class Test_Absolute_spd_to_XYZ(unittest.TestCase):
```
KelSolaar, line 1325: **Minor**

Needs to be added to `__all__`, and I would probably move it after the main test classes.

colour/colorimetry/tests/test_tristimulus_values.py
```diff
@@ -1320,6 +1322,91 @@ def test_raise_exception_sd_to_XYZ_ASTME308(self):
...
+    """
+    Tests methods for absolute colorimetry using k=683
+    """
+
+    def test_sd_to_XYZ_absolute_1nm(self):
```
KelSolaar, line 1330: **Minor**

I would not mind a docstring here please.

colour/colorimetry/tests/test_tristimulus_values.py
```diff
@@ -1320,6 +1322,91 @@ def test_raise_exception_sd_to_XYZ_ASTME308(self):
...
+        for method in methods[3:6]:
+            xyz: np.ndarray = method(spd, k=683)
+            if len(xyz.shape) > 1:
+                xyz = xyz.reshape(3)
+            assert xyz[1] == approx(683, 0.000005), (
+                "1 W @ 555nm should be approximately 683 candela."
+                f" Failed method: {method}"
+            )
+
+    def test_sd_to_XYZ_absolute_5nm(self):
```
KelSolaar, line 1368: **Minor**

Same here!

KelSolaar (commented):
Thanks @tjdcs! I left a few comments, only one to really address.

colour/colorimetry/tests/test_tristimulus_values.py
tjdcs: Will pytest become a hard dev dependency in the future? It has a few nice features. And actually, I used some of them in a previous PR so you may want to check.

KelSolaar: Yes, I have considered it for a while but we will do that in one go, the `conftest.py` file in the `plotting` sub-package is fairly well isolated and more a convenient thing for people using `pytest`.

tjdcs: Resolved in d98a69b8f444f426eec28ed43ba252f7e48a991f

tjdcs:
@KelSolaar  This is ready for another review or merge

colour/recovery/tests/test_jakob2019.py
```diff
@@ -203,32 +203,25 @@ class TestLUT3D_Jakob2019(unittest.TestCase):
     definition unit tests methods.
     """
 
-    def setUp(self):
+    @classmethod
+    def get_cls_lut(self):
```
KelSolaar, line 204: Was there a reason to do that, the `setUp` function should be ran once unless it is ran many times with pytest xdist?

tjdcs: It runs once for every call to every test, not once for every class. Even when I disabled xdist. Which is why each of these tests were taking like 2 minutes to run on a fast PC.

tjdcs: https://docs.python.org/3/library/unittest.html#setupclass-and-teardownclass:~:text=The%20setUp()%20and%20tearDown()%20methods%20allow%20you%20to%20define%20instructions%20that%20will%20be%20executed%20before%20and%20after%20each%20test%20method.%20They%20are%20covered%20in%20more%20detail%20in%20the%20section%20Organizing%20test%20code.

KelSolaar: Right, so why then not simply keeping the method pretty much as is and do something along those lines:

```python
    def setUp(self):
        """Initialise the common tests attributes."""

        if not hasattr(self, "_LUT"):
            self._shape = SPECTRAL_SHAPE_JAKOB2019
            self._cmfs, self._sd_D65 = handle_spectral_arguments(
                shape_default=self._shape
            )
            self._XYZ_D65 = sd_to_XYZ(self._sd_D65)
            self._xy_D65 = XYZ_to_xy(self._XYZ_D65)

            self._RGB_colourspace = RGB_COLOURSPACE_sRGB

            self._temporary_directory = tempfile.mkdtemp()

            self._LUT = LUT3D_Jakob2019()
            self._LUT.generate(self._RGB_colourspace, self._cmfs, self._sd_D65, 5)

            self._path = os.path.join(
                self._temporary_directory, "Test_Jakob2019.coeff"
            )
            self._LUT.write(self._path)
```

tjdcs: I did attempt this. But the self._LUT attribute was not being maintained (state) between test cases. So it needed to be a class method.

tjdcs: I did test this style caching with xdist -n 1 and with no xdist. It doesn't seem to be working for me at all. I'd recommend merging this as is and investigating the test caching stuff at another date. 

This technically saves a small mount of function call time on the other test cases that do not use _Lut. But it's negligible.

KelSolaar: Having looked at that tonight a bit, it is more complicated than expected because each `xdist` worker will need to instantiate the class and call it. Putting some warnings in your current code inside the if:

```
        if not hasattr(self, "_LUT"):
            import time
            import warnings
            warnings.warn(f"Time: {time.time()}")
```

I can see 3 warnings at least:

```
colour/recovery/tests/test_jakob2019.py::TestLUT3D_Jakob2019::test_lightness_scale
  /Users/kelsolaar/Documents/Development/colour-science/colour/colour/recovery/tests/test_jakob2019.py:209: UserWarning: Time: 1665477621.7340338
    warnings.warn(f"Time: {time.time()}")

colour/recovery/tests/test_jakob2019.py::TestLUT3D_Jakob2019::test_coefficients
colour/recovery/tests/test_jakob2019.py::TestLUT3D_Jakob2019::test_LUT3D_Jakob2019
  /Users/kelsolaar/Documents/Development/colour-science/colour/colour/recovery/tests/test_jakob2019.py:209: UserWarning: Time: 1665477621.73266
    warnings.warn(f"Time: {time.time()}")

colour/recovery/tests/test_jakob2019.py::TestLUT3D_Jakob2019::test_size
  /Users/kelsolaar/Documents/Development/colour-science/colour/colour/recovery/tests/test_jakob2019.py:209: UserWarning: Time: 1665477622.012844
    warnings.warn(f"Time: {time.time()}")
```

Those two methods that work perfectly with `Unitest` are called for each test with `xdist`:

```python

    def __new__(cls, *args, **kwargs):
        """
        Initialise the common tests attributes.

        Notes
        -----
        -   We use here the :meth:`unittest.TestCase.__new__` method to define
            the common tests attributes once rather than the typical
            :meth:`unittest.TestCase.setUp` method.
        """

        if not hasattr(cls, "_LUT"):
            import time
            import warnings
            warnings.warn(f"Time: {time.time()}")
            cls._shape = SPECTRAL_SHAPE_JAKOB2019
            cls._cmfs, cls._sd_D65 = handle_spectral_arguments(
                shape_default=cls._shape
            )
            cls._XYZ_D65 = sd_to_XYZ(cls._sd_D65)
            cls._xy_D65 = XYZ_to_xy(cls._XYZ_D65)

            cls._RGB_colourspace = RGB_COLOURSPACE_sRGB

            cls._temporary_directory = tempfile.mkdtemp()

            cls._LUT = LUT3D_Jakob2019()
            cls._LUT.generate(cls._RGB_colourspace, cls._cmfs, cls._sd_D65, 5)

        cls._path = os.path.join(
            cls._temporary_directory, "Test_Jakob2019.coeff"
        )
        cls._LUT.write(cls._path)

        return super(TestLUT3D_Jakob2019, cls).__new__(cls)

    def __del__(self):
        """
        After tests actions.

        Notes
        -----
        -   We use here the :meth:`unittest.TestCase.__del__` method to delete
            the resources rather than the :meth:`unittest.TestCase.tearDown`
            method.
        """

        shutil.rmtree(self._temporary_directory)
```

Digging a bit more: https://pytest-xdist.readthedocs.io/en/latest/how-to.html#making-session-scoped-fixtures-execute-only-once

KelSolaar: Might be the cleanest solution: https://stackoverflow.com/a/59504228

tjdcs: Alternatively we could config xdist to run those tests on the same worker by running all test classes on the same worker. ;) This should be effective for dev use and for CI

tjdcs: That is a nice stack overflow. I just enabled the loaddist scheduler across the entire test suite in the config file. Looks like it's working great. [84af7bb](https://github.com/colour-science/colour/pull/1036/commits/84af7bbf8fc9e2ed84d0663d7b1912a40ca1bc24)

KelSolaar: Nice!

KelSolaar:
LGTM!

Last note: Would you please mind splitting this PR in two, one for the sd to XYZ and another one for the Jakob (2019) stuff. I can do it if you are busy.

Last Question: What's the story with the Visual Studio folder in the `.gitignore` file?

tjdcs:
> LGTM!
> 
> Last note: Would you please mind splitting this PR in two, one for the sd to XYZ and another one for the Jakob (2019) stuff. I can do it if you are busy.
> 
> Last Question: What's the story with the Visual Studio folder in the `.gitignore` file?

Git ignore was updated because I opened the project folder in visual studio and it created a dir for it's config files. 

I'm not exactly sure of the best way to re-write history and separate the commits. I'd like to learn, but I'm not going to have time this week. If you wouldn't mind finishing the merge that would be great!

KelSolaar:
Don't worry, I will do that this weekend! Thanks a lot!

KelSolaar:
Thank you, closing in favour of the #1050 and #1051 split.

PR #1057, opened 2022-10-24
PR: Improve initialization time of "Signal", a core component of "Colour".

tjdcs:
# Summary

This PR implements a performance improvement for `Signal` in `signal.py`. 

The initialization of _function, the private backing attribute for the `function` property, is relatively expensive. This performance issue was discovered when creating an MSDS object from a long list of SpectralDistributions or large ndarrays. The _create_function is invoked frequently when not needed. This PR updates relavent code to always use the function property. the function property has been changed to only invoke _create_function if the _function property has been invalidated (by setting it equal to None). If the function property is never accessed, then the relevant initialization does not take place (lazy initialization)

Under some conditions, this change results in a 90% reduction in Signal initialization run time

# Preflight

**Code Style and Quality**

**Documentation**

### DIFF
--- a/colour/continuous/signal.py
+++ b/colour/continuous/signal.py
@@ -276,7 +276,7 @@ class Signal(AbstractContinuousFunction):
             "extrapolator_kwargs", self._extrapolator_kwargs
         )
 
-        self._create_function()
+        self._function = None
 
     @property
     def dtype(self) -> Type[DTypeFloating]:
@@ -349,7 +349,7 @@ class Signal(AbstractContinuousFunction):
             self._range = np.resize(self._range, value.shape)
 
         self._domain = value
-        self._create_function()
+        self._function = None  # Invalidate internal _function
 
     @property
     def range(self) -> NDArray:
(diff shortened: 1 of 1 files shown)


tjdcs:
Looks like this is well formed now.

tjdcs:
@KelSolaar Other than the issue with static type checking (looks like something related to a new version of pip and an old installation method from a dependency) this is ready to merge.

KelSolaar:
Hey @tjdcs,

Thank you! Two notes:

- The Static Typing issue is because we need to `cast` the return value of the `function` method to a `Callable`:

```python
    @property
    def function(self) -> Callable:
        """
        Getter property for the continuous signal callable.

        Returns
        -------
        Callable
            Continuous signal callable.
        """

        if self._function is None:
            self._create_function()

        return cast(Callable, self._function)
```

- Given that the `_create_function` is only used once now, we should probably move its code inside the `function` method directly.

KelSolaar:
Thank you!

There are some issues with GH so I cannot comment on the changes directly, anyway, the `Callable | None` construct was introduced in [Python 3.10](https://mypy.readthedocs.io/en/stable/kinds_of_types.html#x-y-syntax-for-unions) so we tend to use the longer `Union[Callable, None]` form for now. I will merge and fix in a subsequent commit!

Issue #1063, opened 2022-11-04
[BUG]: "colour.rgb_to_ycbcr" definition can give invalid answers for edge values.


kaihagseth:
### Description

rgb_to_ycbcr can give cb and cr values outside the defined out bit range when outputting to full range. If I feed it with 100% red the function return 1024 for the red channel, which is of course outside the range of 10 bits.

I believe that this is caused by these lines of code:
ycbcr.py:147
```python
    if is_int and not is_legal:
        ranges[3] = 2**bits
```

~~I think the fix would be just to remove these lines, but since it looks so intentional I'll leave it to the maintainer.~~ 

Actually, thinking about it again, the correct solution would be to set ranges[2] to 1.
Rec.2100, which addresses full range on page 10.
https://www.itu.int/dms_pubrec/itu-r/rec/bt/R-REC-BT.2100-2-201807-I!!PDF-E.pdf

Which means my suggestion is that we should make this fix:
ycbcr.py:147
```python
    if is_int and not is_legal:
        ranges[2] = 1
```

### Code for Reproduction
```python
import colour

rgb = (1.0, 0, 0)
y, cb, cr = colour.RGB_to_YCbCr(rgb, in_int=False, in_legal=False, out_bits=10, out_legal=False, out_int=True)
print(y, cb, cr)
# >> 217 395 1024
```

### Environment Information

```shell
>>> import colour; colour.utilities.describe_environment()
===============================================================================
*                                                                             *
*   Interpreter :                                                             *
*       python : 3.9.6 (default, Sep 26 2022, 11:37:49)                       *
*                [Clang 14.0.0 (clang-1400.0.29.202)]                         *
*                                                                             *
*   colour-science.org :                                                      *
*       colour : 0.4.1                                                        *
*                                                                             *
*   Runtime :                                                                 *
*       imageio : 2.22.3                                                      *
*       numpy : 1.23.4                                                        *
*       scipy : 1.9.3                                                         *
*                                                                             *
===============================================================================
defaultdict(<class 'dict'>, {'Interpreter': {'python': '3.9.6 (default, Sep 26 2022, 11:37:49) \n[Clang 14.0.0 (clang-1400.0.29.202)]'}, 'colour-science.org': {'colour': '0.4.1'}, 'Runtime': {'imageio': '2.22.3', 'numpy': '1.23.4', 'scipy': '1.9.3'}})
```

KelSolaar:
[...]

PR #1064, opened 2022-11-09
PR: Modify JFIF full range YCbCr.

nick-shaw:
Changes to address points discussed in #1063

Modifies full range integer Y'CbCr (JFIF) to better match the spec in ITU-T T.871. Also applies to BT.2100.

Adds a `clamp_int` argument to `RGB_to_YCbCr` and `YCbCr_to_RGB` to limit integer output to valid range for `bit_depth`.

The `clamp_int` argument is not added to `YcCbcCrc_to_RGB`. @KelSolaar please add if you think appropriate.


KelSolaar (approved):
LGTM! Thanks @nick-shaw!

PR #1067, opened 2022-11-14
PR: Performance improvements to `colour.SpectralDistribution.wavelengths` property and various definitions in `colour.utilities.array` module.

tjdcs:
# Summary

This PR makes small changes to improve the performance of several critical paths that are found throughout the code base. For example, tsplit and tstack are called very often. Additionally, the `shape` getter of `SpectralDistribution` is called very often in code that is manipulating spectral data. In one example calculating color rendering simulations, calling the .shape getter was responsible for around 10% of total execution time, but is now responsible for less than 0.1% 

The particular examples that found these critical paths are not expected to be representative of all performance cases, none the less nearly all applications of colour-science should see some benefit. 

# Preflight

**Code Style and Quality**

**Documentation**

### DIFF
--- a/colour/colorimetry/spectrum.py
+++ b/colour/colorimetry/spectrum.py
@@ -843,11 +843,19 @@ class SpectralDistribution(Signal):
                 f"minimum interval!"
             )
 
-        return SpectralShape(
-            min(self.wavelengths),
-            max(self.wavelengths),
-            min(wavelengths_interval),
-        )
+        # If wavelegths is sorted, then go fast. Otherwise compute the shape.
+        if np.all(self.wavelengths[:-1] <= self.wavelengths[1:]):
+            return SpectralShape(
+                self.wavelengths[0],
+                self.wavelengths[-1],
+                min(wavelengths_interval),
+            )
+        else:
+            return SpectralShape(
+                min(self.wavelengths),
+                max(self.wavelengths),
+                min(wavelengths_interval),
+            )
 
     def interpolate(
         self,
--- a/colour/colorimetry/spectrum.py
+++ b/colour/colorimetry/spectrum.py
@@ -843,17 +843,17 @@ class SpectralDistribution(Signal):
                 f"minimum interval!"
             )
 
-        # If wavelegths is sorted, then go fast. Otherwise compute the shape.
-        if np.all(self.wavelengths[:-1] <= self.wavelengths[1:]):
+        # If wavelengths is sorted, then go fast. Otherwise compute the shape.
+        if np.all(self.domain[:-1] <= self.domain[1:]):
             return SpectralShape(
-                self.wavelengths[0],
-                self.wavelengths[-1],
+                self.domain[0],
+                self.domain[-1],
                 min(wavelengths_interval),
             )
         else:
             return SpectralShape(
-                min(self.wavelengths),
-                max(self.wavelengths),
+                min(self.domain),
+                max(self.domain),
                 min(wavelengths_interval),
             )
 
--- a/colour/utilities/array.py
+++ b/colour/utilities/array.py
@@ -928,7 +928,7 @@ def get_domain_range_scale() -> Union[
 
 def set_domain_range_scale(
     scale: Union[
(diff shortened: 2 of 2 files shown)


colour/utilities/array.py
```diff
@@ -2031,6 +2031,8 @@ def tstack(
     dtype = cast(Type[DTypeFloating], optional(dtype, DEFAULT_FLOAT_DTYPE))
 
     a = as_array(a, dtype)
+    if a.ndim <= 2:
+        return a.T
```
KelSolaar, line 2035: `np.transpose(a)` makes me happier :)

colour/utilities/array.py
```diff
@@ -928,7 +928,7 @@ def get_domain_range_scale() -> Union[
 
 def set_domain_range_scale(
     scale: Union[
-        Literal["ignore", "reference" "Ignore", "Reference", "1", "100"], str
+        Literal["ignore", "reference", "Ignore", "Reference", "1", "100"], str
```
KelSolaar, line 931: Good catch!

colour/utilities/array.py
```diff
@@ -1027,7 +1027,7 @@ class domain_range_scale:
     def __init__(
         self,
         scale: Union[
-            Literal["ignore", "reference" "Ignore", "Reference", "1", "100"],
+            Literal["ignore", "reference", "Ignore", "Reference", "1", "100"],
```
KelSolaar, line 1030: Ditto!

colour/utilities/array.py
```diff
@@ -2096,7 +2096,13 @@ def tsplit(
 
     a = as_array(a, dtype)
 
-    return np.array([a[..., x] for x in range(a.shape[-1])])
+    if a.ndim <= 2:
+        return a.T
```
KelSolaar, line 2102: np.transpose(a) makes me happier :)

colour/colorimetry/spectrum.py
```diff
@@ -843,11 +843,19 @@ def shape(self) -> SpectralShape:
                 f"minimum interval!"
             )
 
-        return SpectralShape(
-            min(self.wavelengths),
-            max(self.wavelengths),
-            min(wavelengths_interval),
-        )
+        # If wavelegths is sorted, then go fast. Otherwise compute the shape.
+        if np.all(self.wavelengths[:-1] <= self.wavelengths[1:]):
```
KelSolaar, line 847: At the point where we try to get some speed, should not we also store `self.wavelengths` into an intermediate variable so that we do not access and copy it two times (see `colour.Signal.range` property)?

What about this: `np.all(np.diff(wavelengths) >= 0)` ?

colour/colorimetry/spectrum.py
```diff
@@ -843,11 +843,19 @@ def shape(self) -> SpectralShape:
                 f"minimum interval!"
             )
 
-        return SpectralShape(
-            min(self.wavelengths),
-            max(self.wavelengths),
-            min(wavelengths_interval),
-        )
+        # If wavelegths is sorted, then go fast. Otherwise compute the shape.
```
KelSolaar, line 846: `wavelegths` --> `wavelengths`

KelSolaar (commented):
Thank you! I left a few minor notes but LGTM otherwise!

colour/colorimetry/spectrum.py
tjdcs: I tested diff and it's actually slower than this. I will check memcopys though.

colour/utilities/array.py
tjdcs: a.T about 25% faster.  if the difference was smaller, then it would be worth it to trade off for clarity. but this function is called a lot

colour/colorimetry/spectrum.py
tjdcs: Double checked. I couldn't find a way to eliminate any further copies.

colour/colorimetry/spectrum.py
tjdcs: thx

PR #1117, opened 2023-03-08
Add feature for calculating continuous `spectral_similarity_index` values

tjdcs:
# Summary

This adds the functionality for computing a continuous spectral similarity index values, which can be useful for optimization problems. 

Additionally there is a small code change in array.py which resulted in meaningful performance improvements with real problems. 

# Preflight

**Code Style and Quality**

- [ ] Unit tests have been implemented and passed.
- [ ] Pre-commit hooks have been run and passed.
- [ ] New transformations have been added to the *Automatic Colour Conversion Graph*.
- [ ] New transformations have been exported to the relevant namespaces, e.g. `colour`, `colour.models`.

**Documentation**

- [ ] New features are documented along with examples if relevant.


tjdcs:
Regarding the change in array.py, I did test some alternatives that explicitly use the transpose function. It seems that these were all slower and the shorthand .T property, probably just sets a flag in the returned data to lazy transpose. In an optimization use case I was working on, this did result in a valuable performance improvement. Strange as it may look.

tjdcs:
> Regarding the change in array.py, I did test some alternatives that explicitly use the transpose function. It seems that these were all slower and the shorthand .T property, probably just sets a flag in the returned data to lazy transpose. In an optimization use case I was working on, this did result in a valuable performance improvement. Strange as it may look.

I realized I had dead code there and re-investigated. Can't replicate whatever was making me thing this, so updated to `np.transpose` with no dead code.

colour/quality/ssi.py
```diff
@@ -56,6 +58,8 @@ def spectral_similarity_index(
         Test spectral distribution.
     sd_reference
         Reference spectral distribution.
+    continuous
```
KelSolaar, line 61: Should we call it `round_output`, `rounded_output` rather than `continuous`? Seems more explicit!

tjdcs: Done

.gitignore
```diff
@@ -21,4 +21,4 @@ docs/_static/Plotting_*.png
 docs/_static/Tutorial_*.png
 docs/generated
 poetry.lock
-references
+references
```
KelSolaar, line 24:
```suggestion
references

```

colour/utilities/array.py
```diff
@@ -2044,7 +2044,6 @@ def tstack(
     dtype = optional(dtype, DEFAULT_FLOAT_DTYPE)
 
     a = as_array(a, dtype)
-
     if a.ndim <= 2:
```
KelSolaar, line 2047:
```suggestion
    if a.ndim <= 2:
```

KelSolaar:
Thank you!

PR #1151, opened 2023-05-13
PR: Implement support for better `ndarray` copy management in `colour.continuous.Signal` class.

KelSolaar:
# Summary

This PR implements support for better `ndarray` copy management in `colour.continuous.Signal` class enabling most internal operations to be performed without a copy of the data and thus enabling better performance.

The newly added `colour.utilities.ndarray_copy_enable` context manager / decorator should also be useful for other areas touched by #1120. I purposely left them untouched in this particular PR.

# Preflight

**Code Style and Quality**

**Documentation**


colour/utilities/tests/test_array.py
```diff
@@ -1116,6 +1124,95 @@ def test_from_range_int(self):
...
+        """Test :func:`colour.utilities.array.ndarray_copy` definition."""
+
+        a = np.linspace(0, 1, 10)
+        with ndarray_copy_enable(True):
+            self.assertNotEqual(id(ndarray_copy(a)), id(a))
+
+        with ndarray_copy_enable(False):
+            self.assertEqual(id(ndarray_copy(a)), id(a))
+
+
```
tjdcs, line 1215: We should test the other decorating modes as well. This leaves the function decorator untested.

colour/utilities/array.py
```diff
@@ -1788,6 +1792,142 @@ def from_range_int(
     return a
 
 
+_NDARRAY_COPY_ENABLED: bool = True
```
tjdcs, line 1795: I wonder if there is something we should do (long term, not this PR) to manage the number of globals Colour creates. A lot of the caches are also globals. Users might not want to see these globals if they are trying to debug client applications, or at least it could be clearer that they are colour-"globals"

tjdcs (changes requested):
Testing changes requested.

tjdcs (changes requested):
Testing changes requested.

colour/continuous/signal.py
```diff
@@ -515,6 +517,7 @@ def extrapolator_kwargs(self, value: dict):
         self._function = None  # Invalidate the underlying continuous function.
 
     @property
+    @ndarray_copy_enable(False)
```
tjdcs, line 520: It's not clear to me why these decorators are littered around... I can't see where they are actually affecting any behavior. It looks like it is a promise that this function is copy-less safe (no-side effects) but that guarantee or promise didn't exist before and doesn't appear to be adding value to the function signature here... Are we going to end up with `@nd_copy_enable(False)` functions everywhere? 

Minor concern about long term code litter.

colour/utilities/tests/test_array.py
KelSolaar: The decorator is tested here: https://github.com/colour-science/colour/pull/1151/files/27f6f4b79513ab2f00a2dbd09e15ecead8b47e91#diff-a99802bce9e534d4136fb2c2f7d425fdc96e4f19a6505769437c39512f2a983dR1182

colour/utilities/array.py
KelSolaar: This is a good point yeah, it should be easy to do!

colour/continuous/signal.py
KelSolaar: This is a good concern, and one would certainly have to know what is going on here, i.e. that it disable the `domain` and `range` getter copies. We could add some comments to make it understandable. I don't think there is a need to have the decorators anywhere else than where they are.

An alternative is to have getters that are without copy, e.g. `domain_no_copy` or alike.

KelSolaar: I just did a small pass using directly the protected attributes thus removed most of the decorated calls.

tjdcs (approved):
Looks ready to me! Testing went well. 

Will refactor performance branch on this... eventually.

PR #1153, opened 2023-05-19
PR: Remove ability to pass illuminant in "CIE XYZ" to "CIE xyY" related transformations.

KelSolaar:
# Summary

This one has been itching me for a while, it is a very old behaviour that is very strongly opinionated but not recommended by CIE during the computation of chromaticity coordinates: When the tristimulus values are null, we do set the chromaticity coordinates to the passed illuminant ones and the luminance to zero. This is what Bruce Lindbloom is doing. It is quite useful as it means that the achromatic colours are all on the same axis but I reckon that this is a behaviour that should be controlled by the user post-conversion. We do not do that with any other space for that matter.

image

# Preflight

**Code Style and Quality**

- [] Pyright static checking has been run and passed.

**Documentation**


KelSolaar:
Merging this one, as it improves consistency across the board, especially with regard to the few other space that do projective transformations.

PR #1174, opened 2023-06-24
Fix Poetry package resolution when using local versions of colour-science in poetry projects

tjdcs:
Poetry complains: "The dependency name for colour-science does not match the actual package's name: colour" when using a path dependency in local projects. This is useful when a local project is dependent on a local version of color (like adding colour as a submodule dependency to a project)

Example:
```
[tool.poetry.dependencies]
python = ">=3.11,<3.12"
colour-science = {path = "submodules/colour-science", develop = true}
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,8 @@
 [tool.poetry]
-name = "colour"
+name = "colour-science"
+packages = [
+    { include = "colour" },
+]
 version = "0.4.2"
 description = "Colour Science for Python"
 license = "BSD-3-Clause"


KelSolaar:
I will need to look at this one in details, there are complications with the Pypi release and the `tasks.py` file is doing some dirty things because of that. Maybe those are not needed anymore with recent versions of `Poetry` but in the past you could not have a package mismatching the project name, it was basically breaking the build.

tjdcs:
I am deploying colour-specio with this type of config. But it is a much less complicated project.

tjdcs:
Specific example / use case for a "workbench" base project. 

https://github.com/tjdcs/colour-workbench

KelSolaar:
@tjdcs : Was that one closed on purpose?

tjdcs:
Yes, but because I wanted to move the branch to a better branch folder. I will re-open

PR #1209, opened 2023-10-19
Small plotting features

tjdcs:


colour/plotting/common.py
```diff
@@ -494,7 +494,13 @@ def artist(**kwargs: KwargsArtist | Any) -> Tuple[Figure, Axes]:
 
         return figure, figure.gca()
     else:
-        return plt.gcf(), cast(Axes, axes)
+        axes = cast(Axes, axes)
+        figure = axes.figure
+
+        while isinstance(figure, SubFigure):
```
KelSolaar, line 500: Why are we using `while` here?

colour/plotting/common.py
```diff
@@ -29,25 +29,19 @@
...
 import matplotlib.ticker
 import numpy as np
-from cycler import cycler
-from dataclasses import dataclass, field
-from functools import partial
-from matplotlib.axes import Axes
-from matplotlib.colors import LinearSegmentedColormap
-from matplotlib.figure import Figure
-from matplotlib.patches import Patch
-from mpl_toolkits.mplot3d.axes3d import Axes3D
```
KelSolaar, line 43: Let's try to not reorder the imports automatically, it looks like `isort` work here and it is not doing a great job. We follow PEP8: https://peps.python.org/pep-0008/#imports

> Imports are always put at the top of the file, just after any module comments and docstrings, and before module globals and constants.
> Imports should be grouped in the following order:
> 
> Standard library imports.
> Related third party imports.
> Local application/library specific imports.

colour/plotting/common.py
```diff
@@ -701,7 +711,7 @@ def render(
         figure.savefig(settings.filename)
 
     if settings.show:
-        plt.show()
+        plt.show(block=settings.block)
```
KelSolaar, line 714: Nice!

KelSolaar (changes requested):
LGTM besides the imports reorder! See my comment above.

colour/plotting/common.py
tjdcs: Yeah, it became clear to me that this was a way bigger tool / dev change than I was thinking originally. Removed in this PR. 

But I want to revisit this in the future. It's reading - pleasing when my auto-completed imports get put in the correct place by a tool. Ruff has some capability to do this (but I'm not sure about configurability) and isort can probably be made to match pep8.

colour/plotting/common.py
tjdcs: When I first read the docs I thought the get figure might sometimes return a parent subfigure

tjdcs:
Addressed all comments @KelSolaar

KelSolaar:
PR is still draft @tjdcs, not sure if it is intended!

KelSolaar:
`/merged`

PR #1219, opened 2023-10-29
Add is_within_visible_xy implementation

tjdcs:
Add an xy implementation to is_within_visible_spectrum for brightness invariant checking.


KelSolaar:
I think that I would prefer a dedicated separate public definition instead of magic based on the shape. What about something like renaming the current one to `is_within_visible_spectrum_volume` and the `xy` one `is_within_visible_spectrum_area`?

Cheers,

PR #1268, opened 2024-05-29
PR: Improve performance of SpragueInterpolator

tjdcs:
`SpragueInterpolator` is the default interpolator for `SpectralDistribution` making it a highly frequented code path. This PR greatly improves the performance of interpolation by using vectorized code and numpy.

Other interpolators could be similarly sped up but they are lower priority.

### DIFF
--- a/colour/algebra/interpolation.py
+++ b/colour/algebra/interpolation.py
@@ -1204,9 +1204,7 @@ class SpragueInterpolator:
         # Fancy vector code here... use underlying numpy structures to accelerate
         # parts of the linear algebra.
 
-        y = r[i] + np.einsum(
-            "ab,ab->b", a.reshape(5, -1), X ** np.arange(1, 6).reshape(-1, 1)
-        )
+        y = r[i] + (a.reshape(5, -1) * X ** np.arange(1, 6).reshape(-1, 1)).sum(axis=0)
         if y.size == 1:
             return y[0]
         return y
--- a/colour/algebra/interpolation.py
+++ b/colour/algebra/interpolation.py
@@ -1127,42 +1127,14 @@ class SpragueInterpolator:
 
         self._y = value
 
-        yp1 = np.ravel(
-            (
-                np.dot(
-                    self.SPRAGUE_C_COEFFICIENTS[0],
-                    np.reshape(np.array(value[0:6]), (6, 1)),
-                )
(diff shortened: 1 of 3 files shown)


tjdcs:
Two tests are failing

```txt
=========================== short test summary info ============================
FAILED colour/colorimetry/generation.py::colour.colorimetry.generation.sd_gaussian
FAILED colour/colorimetry/generation.py::colour.colorimetry.generation.sd_gaussian_fwhm
===== 2 failed, 3464 passed, 11 skipped, 190 warnings in 185.65s (0:03:05) =====
```
However, my local environment passes these... :/ sneaky. Will investigate.

tjdcs:
@KelSolaar the failing tests are  doctest(s) in `generation.py`. Do you have a prefered resolution? Should I add ellipses... or round... or something else? Seems like the explicit tests are passing within the tolerance we expect / use.

KelSolaar:
Hey @tjdcs,

Thanks a lot! Yeah ellipsis would be absolutely fine here.

Cheers,

Thomas

colour/algebra/interpolation.py
```diff
@@ -1217,36 +1189,26 @@ def _evaluate(self, x: NDArrayFloat) -> NDArrayFloat:
...
+            )
+        )
+        a = np.dot(w_s, r_s) / 24
+
+        # Fancy vector code here... use underlying numpy structures to accelerate
+        # parts of the linear algebra.
 
+        y = r[i] + np.einsum(
+            "ab,ab->b", a.reshape(5, -1), X ** np.arange(1, 6).reshape(-1, 1)
+        )
```
KelSolaar, line 1209: With the desire to move toward data-array API compliance and because Einstein Sum is not planned to be supported, we should try to rewrite this using lower level APIs: https://data-apis.org/array-api/latest/extensions/linear_algebra_functions.html

KelSolaar (commented):
LGTM besides the `einsum` usage, I think we need to get rid of the other usages in the codebase.

colour/algebra/interpolation.py
KelSolaar: I removed all usages of `np.einsum` in `develop`.

tjdcs:
@KelSolaar looks like these are all passing now. Should be g2g when the last few tasks finish.

Issue #1273, opened 2024-06-17
[FEATURE]: Support numpy 2.x


paulschreiber:
### Description

Numpy 2.0 was released. Please add support for Numpy 2.x

KelSolaar:
Hello @paulschreiber,

This is on the TODO list! That being said, we are having some issues with Scipy that currently force us to forbid [1.13.x and lower versions do not support Numpy 2](https://github.com/scipy/scipy/blob/4edfcaa3ce8a387450b6efce968572def71be089/pyproject.toml#L24): https://github.com/scipy/scipy/issues/20613

KelSolaar:
This was merged in `develop`.

FredeJ:
IMO This is a big enough thing to warrant a release on its own, as it allows everyone to update to numpy 2 and start using it. Any chance to see a v0.4.5 release featuring just what's available now, and bumping the rest of the v0.4.5 milestone to a v0.4.6 release? :)

More specifically - this library is currently the only thing holding us back from upgrading to numpy 2, so I'd love to see if there's anything I could do to help out in that process :)

KelSolaar:
Hello @FredeJ,

Thanks, this is dully noted and we will work towards getting a release up. I'm engaged on other projects that have deadlines in coming weeks and the plan was to work on releasing Colour after this.

FredeJ:
Thanks a lot, @KelSolaar! That's obviously totally fair, and it's very nice to have a rough estimate :)

PR #1276, opened 2024-07-01
PR: Numpy 2 Compatibility: Replace `np.float_` w/ `np.double`.

trougnouf:
# Summary

Replace np.float_ with np.double for numpy 2.0 compatibility. This allows importing the colour module without error again. ( #1273 )


KelSolaar:
Thanks @trougnouf!

Issue #610, opened 2020-07-18
Implement support for "HyAB" large colour difference metric.


KelSolaar:
References
---
- https://onlinelibrary.wiley.com/doi/abs/10.1002/col.22451

NekoAlosama:
Here's a copy of the PDF provided by one of the authors (Mark D. Fairchild) on their website: http://markfairchild.org/PDFs/PAP40.pdf

NekoAlosama:
@KelSolaar 
Here's a copy of the research paper provided by one of the authors (Mark D. Fairchild) on their website: http://markfairchild.org/PDFs/PAP40.pdf

NekoAlosama:
HyAB is just the combination of a city-block expression for L* and the Euclidean distance with a* and b*.
So alike `abs(base.L*- test.L*) + math.dist([base.a*, base.b*], [test.a*, test.b*])` with L\*a\*b\* colors 'base' and 'test'.

KelSolaar:
Thanks @NekoAlosama! I have the papers but did not find the time so far to implement the metric! If you want to give it a go, you are very much welcome do to so!

PR #1305, opened 2024-10-24
PR: Implement support for *HyAB*  large color difference.

mesvam:
# Summary

Implement HyAB color difference and resolves https://github.com/colour-science/colour/issues/610

# Preflight

**Code Style and Quality**

- [ ] New transformations have been added to the _Automatic Colour Conversion Graph_.

**Documentation**


mesvam:
I think I got the implementation correct, but could somebody double check the documentation, citation, formatting, etc.? I'm not sure I got all the tests to run correctly.

colour/difference/delta_e.py
```diff
@@ -560,3 +566,56 @@ def delta_E_ITP(ICtCp_1: ArrayLike, ICtCp_2: ArrayLike) -> NDArrayFloat:
...
+
+    References
+    ----------
+    :cite:`Abasi2020`
+
+    Examples
+    --------
+    >>> Lab_1 = np.array([25.0, 8.0, -14.0])
+    >>> Lab_2 = np.array([75.0, -35.0, 16.0])
+    >>> delta_E_CIE2000(Lab_1, Lab_2)  # doctest: +ELLIPSIS
```
KelSolaar, line 615: **Major**

The unit tests are likely failing because of this one, it should be calling `delta_E_HyAB` and not `delta_E_CIE200`.

colour/difference/delta_e.py
```diff
@@ -560,3 +566,56 @@ def delta_E_ITP(ICtCp_1: ArrayLike, ICtCp_2: ArrayLike) -> NDArrayFloat:
...
+    102.4309069...
+    """
+    dLab = to_domain_100(Lab_2) - to_domain_100(Lab_1)
+    dL, da, db = np.moveaxis(dLab, -1, 0)
```
KelSolaar, line 619: **Minor**

I would suggest using `tsplit` for consistency with the remaining of the codebase.

colour/difference/delta_e.py
```diff
@@ -560,3 +566,56 @@ def delta_E_ITP(ICtCp_1: ArrayLike, ICtCp_2: ArrayLike) -> NDArrayFloat:
...
+    :cite:`Abasi2020`
+
+    Examples
+    --------
+    >>> Lab_1 = np.array([25.0, 8.0, -14.0])
+    >>> Lab_2 = np.array([75.0, -35.0, 16.0])
+    >>> delta_E_CIE2000(Lab_1, Lab_2)  # doctest: +ELLIPSIS
+    102.4309069...
+    """
+    dLab = to_domain_100(Lab_2) - to_domain_100(Lab_1)
```
KelSolaar, line 618: **Major**

The metric is actually expressed as $a^\*_1 - a^\*_2$ not $a^\*_2- a^\*_1$:

image

KelSolaar (changes requested):
Thank you @mesvam,

I added a few important comments, what about also adding `cbLAB`? Would be great!

Cheers,

Thomas

colour/difference/delta_e.py
mesvam: Thanks!

colour/difference/delta_e.py
mesvam: Thanks!

colour/difference/delta_e.py
mesvam: done

mesvam:
I just noticed there are functions for `manhattan_distance` and `euclidean_distance` in `colour.algebra`. Would you prefer I use those functions to implement these difference metrics instead?

mesvam:
>  what about also adding `cbLAB`? Would be great!

I don't mind adding cbLAB if you like, as well as some of the other metrics in that paper, but the authors recommended HyAB and HyCH as the top candidates. We can also add all four formulas, but maybe that's too much or too confusing.

Also, would you prefer that in a separate PR, or just tack it onto this one?

mesvam:
Ok, I'm stumped. How do I pass Codacy? I can't seem to find the error messages

KelSolaar:
> I just noticed there are functions for `manhattan_distance` and `euclidean_distance` in `colour.algebra`. Would you prefer I use those functions to implement these difference metrics instead?

If you can wherever possible, this would be fantastic!

I would probably try to get *HyCH* actually if you please don't mind:

> The HyAB has the best consistency with CIELAB for
all categories of the second dataset. The HyCH has a good
consistency with CIEDE2000 for the second dataset as
well. So, the HyAB and HyCH could be used for small to
large color difference calculation. However, the HyAB
has a slightly better performance and the advantage of
simpler computation.
It is thus concluded that, for very large color differ-
ences, the effect of lightness difference should be consid-
ered perceptually separable, and the HyAB color
difference formula is the most practical and accurate
solution.

> Ok, I'm stumped. How do I pass Codacy? I can't seem to find the error messages

Do not worry, it is all good atm!

KelSolaar (approved):
Thank you! Merging this one!

PR #1320, opened 2025-01-03
PR: Add default `shape` argument to `colour.quality.cfi2017.sd_reference_illuminant` definition.

tjdcs:
I frequently use this function elsewhere, and having a default shape is convenient. :)

### DIFF
--- a/colour/quality/cfi2017.py
+++ b/colour/quality/cfi2017.py
@@ -42,7 +42,8 @@ from colour.colorimetry import (
 if typing.TYPE_CHECKING:
     from colour.hints import ArrayLike, List, Literal, Tuple
 
-from colour.hints import NDArrayFloat, cast
+from colour.colorimetry.spectrum import SPECTRAL_SHAPE_DEFAULT
+from colour.hints import ArrayLike, List, NDArrayFloat, Tuple, cast
 from colour.models import JMh_CIECAM02_to_CAM02UCS, UCS_to_uv, XYZ_to_UCS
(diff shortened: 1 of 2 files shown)


colour/quality/tests/test_cfi2017.py
```diff
@@ -924,6 +925,14 @@ class TestSdReferenceIlluminant:
...
     """
 
+    def test_default_args(self) -> None:
+        """Test :func:`color.quality.CIE2017.sd_reference_illuminant` for
+        default shape argument.
+        """
+
+        sd = sd_reference_illuminant(5421)
+        assert sd.shape == SPECTRAL_SHAPE_DEFAULT
+
```
KelSolaar, line 935: **Minor**

It is fine to have this test in the `test_sd_reference_illuminant` method. We are not doing "true" unit tests.

colour/quality/cfi2017.py
```diff
@@ -66,16 +67,16 @@
...
+    "CCT_reference_illuminant",
     "ColourRendering_Specification_CIE2017",
+    "DataColorimetry_TCS_CIE2017",
     "colour_fidelity_index_CIE2017",
+    "delta_E_to_R_f",
     "load_TCS_CIE2017",
-    "CCT_reference_illuminant",
     "sd_reference_illuminant",
     "tcs_colorimetry_data",
-    "delta_E_to_R_f",
```
KelSolaar, line 78: **Question**

Why addition/removal/order change of the objects?

thomasmansencal (changes requested):
Thanks, this is useful!

I left two small comments.

PR #1321, opened 2025-01-03
Add some in-situ documentation to pre commit config

tjdcs:
This helps me see what the current pre-commit config tools are when I happen to need to look at this again in several months. 

‼️ Ahead of PR #1320

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,3 +1,10 @@
+# The pre-commit steps used by colour are:
+# * codespell
+# * isort
+# * ruff-format
+# * ruff --fix
+# * blacken-docs
+# * prettier (with specific fixes)
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: "v5.0.0"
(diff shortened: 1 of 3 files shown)


thomasmansencal (commented):
We are running more hooks than this list, technically:

- check-added-large-files
- check-case-conflict
- check-merge-conflict
- check-symlinks
- check-yaml
- debug-statements
- end-of-file-fixer
- mixed-line-ending
- name-tests-test
- requirements-txt-fixer
- trailing-whitespace
- codespell
- isort
- ruff-format
- ruff
- blacken-docs
- prettier
- rst-backticks
- rst-directive-colons
- rst-inline-touching-normal

I'm not super excited by this because we have that configuration in all the repositories that we distribute on Pypi and if we change it here, we need to change it consistently with the other ones so that patching is easier:

- https://github.com/colour-science/colour-demosaicing/blob/develop/.pre-commit-config.yaml
- https://github.com/colour-science/colour-hdri/blob/develop/.pre-commit-config.yaml
- https://github.com/colour-science/colour-datasets/blob/develop/.pre-commit-config.yaml
- https://github.com/colour-science/colour-checker-detection/blob/develop/.pre-commit-config.yaml
- https://github.com/colour-science/colour-visuals/blob/develop/.pre-commit-config.yaml
- https://github.com/colour-science/colour-demosaicing/blob/develop/.pre-commit-config.yaml
- https://github.com/colour-science/colour-science.org/blob/master/.pre-commit-config.yaml

and I might have forgotten a few others.

This is generally true for most of our config files, if we do one change, we likely want to propagate it to all the other repositories.

PR #1322, opened 2025-01-06
PR: Add `colour.SpectralShape.from_array` class method (constructor).

tjdcs:
This adds a utility constructor for constructing a SpectralShape directly from a list of values.

Not sure if this is already available elsewhere, but if it is then perhaps this organization (placing the method inside the definition of SpectralShape) is better?

‼️Ahead of #1321

### DIFF
--- a/colour/colorimetry/tests/test_spectrum.py
+++ b/colour/colorimetry/tests/test_spectrum.py
@@ -1417,6 +1417,21 @@ class TestSpectralShape:
             np.arange(0, 10 + 0.1, 0.1),
         )
 
+    def test_from_data(self) -> None:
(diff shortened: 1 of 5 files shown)


colour/colorimetry/tests/test_spectrum.py
```diff
@@ -1417,6 +1417,21 @@ def test_range(self) -> None:
             np.arange(0, 10 + 0.1, 0.1),
         )
 
+    def test_from_data(self) -> None:
```
KelSolaar, line 1420: **Major**

`test_from_data` --> `test_from_array`/`test_from_wavelengths`

thomasmansencal (changes requested):
Thanks, great idea!

I left a few minor comments that would be good to address.

PR #1342, opened 2025-04-06
PR: Add `scipy-stubs` as development dependency.

jorenham:
# Summary

This adds [scipy-stubs](https://github.com/scipy/scipy-stubs) as a development dependency. By doing so, 58 new pyright errors popped up. There were 18 that were caused by two bugs and two other issues in `scipy-stubs`, and I've fixed them. See https://github.com/scipy/scipy-stubs/issues/495 for details. 

The other 40 errors turned out to be typing issues in `colour` itself. I have included their fixes in this PR.

~I also noticed that an additional error pops up when the latest `numpy==2.2.4` release is installed (currently pinned at `2.2.0`). I've addressed this in https://github.com/numpy/numpy/pull/28660, and the fix will be included in the upcoming `2.2.5` (and `2.3.0`) NumPy release(s).~ (resolved in [numpy 2.2.5](https://github.com/numpy/numpy/releases/tag/v2.2.5))

This additionally improves the static-typing compatibility for different versions of pyright, numpy, scipy, and scipy-stubs.

# Preflight

**Code Style and Quality**

### DIFF
--- a/colour/algebra/interpolation.py
+++ b/colour/algebra/interpolation.py
@@ -1292,7 +1292,7 @@ class PchipInterpolator(scipy.interpolate.PchipInterpolator):
     """
 
     def __init__(self, x: ArrayLike, y: ArrayLike, *args: Any, **kwargs: Any) -> None:
-        super().__init__(x, y, *args, **kwargs)
+        super().__init__(as_float_array(x), as_float_array(y), *args, **kwargs)
 
         self._y: NDArrayFloat = as_float_array(y)
 
--- a/colour/appearance/rlab.py
+++ b/colour/appearance/rlab.py
@@ -273,6 +273,7 @@ b=-52.6142956...)
     M = np.matmul(np.matmul(MATRIX_R, row_as_diagonal(LMS_a_L)), MATRIX_XYZ_TO_HPE)
     XYZ_ref = vecmul(M, XYZ)
 
+    Y_ref: NDArrayFloat
     X_ref, Y_ref, Z_ref = tsplit(XYZ_ref)
 
     # Computing the correlate of *Lightness* :math:`L^R`.
--- a/colour/characterisation/aces_it.py
+++ b/colour/characterisation/aces_it.py
@@ -831,9 +831,7 @@ finaliser_function at 0x...>)
 
     x_0 = as_float_array([1, 0, 0, 1, 0, 0])
 
-    def objective_function(
-        M: ArrayLike, RGB: ArrayLike, Jab: ArrayLike
-    ) -> DTypeFloat:
+    def objective_function(M: ArrayLike, RGB: ArrayLike, Jab: ArrayLike) -> DTypeFloat:
         """:math:`J_za_zb_z` colourspace based objective function."""
 
         M = finaliser_function(M)
@@ -902,9 +900,7 @@ finaliser_function at 0x...>)
 
     x_0 = as_float_array([1, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 1])
 
-    def objective_function(
-        M: ArrayLike, RGB: ArrayLike, Jab: ArrayLike
-    ) -> DTypeFloat:
+    def objective_function(M: ArrayLike, RGB: ArrayLike, Jab: ArrayLike) -> DTypeFloat:
         """*Oklab* colourspace based objective function."""
 
         M = finaliser_function(M)
--- a/colour/characterisation/aces_it.py
+++ b/colour/characterisation/aces_it.py
@@ -87,6 +87,7 @@ if typing.TYPE_CHECKING:
         Any,
         ArrayLike,
         Callable,
+        DTypeFloat,
         Literal,
         LiteralChromaticAdaptationTransform,
         Mapping,
@@ -768,7 +769,7 @@ def optimisation_factory_rawtoaces_v1() -> (
 
     def objective_function(
         M: NDArrayFloat, RGB: NDArrayFloat, Lab: NDArrayFloat
-    ) -> NDArrayFloat:
+    ) -> DTypeFloat:
         """Objective function according to *RAW to ACES* v1."""
 
         M = finaliser_function(M)
@@ -832,7 +833,7 @@ finaliser_function at 0x...>)
 
     def objective_function(
         M: ArrayLike, RGB: ArrayLike, Jab: ArrayLike
-    ) -> NDArrayFloat:
+    ) -> DTypeFloat:
         """:math:`J_za_zb_z` colourspace based objective function."""
 
         M = finaliser_function(M)
@@ -903,7 +904,7 @@ finaliser_function at 0x...>)
 
     def objective_function(
         M: ArrayLike, RGB: ArrayLike, Jab: ArrayLike
-    ) -> NDArrayFloat:
+    ) -> DTypeFloat:
         """*Oklab* colourspace based objective function."""
 
         M = finaliser_function(M)
@@ -1124,7 +1125,8 @@ def matrix_idt(
         XYZ_to_optimization_colour_model,
         finaliser_function,
     ) = optimisation_factory()
-    optimisation_settings = {
+
+    optimisation_settings: dict[str, Any] = {
         "method": "BFGS",
         "jac": "2-point",
     }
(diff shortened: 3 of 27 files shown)


jorenham:
I have analyzed the typing errors in https://github.com/scipy/scipy-stubs/issues/495#issuecomment-2781190107. Through this, I found three bugs in `scipy-stubs` that I'll soon fix. 
But the majority of the typing errors are true positives, and should probably be fixed in within colour itself. Unless anyone objects, I'd be willing to do include those fixes in this PR.

KelSolaar:
Thanks @jorenham, sounds good with me!

jorenham:
All errors are fixed when the `master` branch of `scipy-stubs` is installed. I could either (subtly) work around the remaining errors for now, or publish a `scipy-stubs` release and upgrade `scipy` and `scipy-stubs` to `1.15.2` and `1.15.2.2`, respectively. What would have your preference?

colour/algebra/interpolation.py
```diff
@@ -1292,7 +1292,7 @@ class PchipInterpolator(scipy.interpolate.PchipInterpolator):
     """
 
     def __init__(self, x: ArrayLike, y: ArrayLike, *args: Any, **kwargs: Any) -> None:
-        super().__init__(x, y, *args, **kwargs)
+        super().__init__(as_float_array(x), as_float_array(y), *args, **kwargs)
```
jorenham, line 1295: `typing.cast` would also be an option, but strictly speaking I'd have to cast it to `Any` for it to be theoretically allowed. Downcasting this correctly would require recreating the `ToFloatND` or `ToFloat1D` types from `optype.numpy` ([docs](https://github.com/jorenham/optype#array-likes)) here, which would be rather messy.

colour/appearance/rlab.py
```diff
@@ -273,6 +273,7 @@ def XYZ_to_RLAB(
     M = np.matmul(np.matmul(MATRIX_R, row_as_diagonal(LMS_a_L)), MATRIX_XYZ_TO_HPE)
     XYZ_ref = vecmul(M, XYZ)
 
+    Y_ref: NDArrayFloat
```
jorenham, line 276: This ensures that the right `@overload` of `spow` is selected, so that `LR` will be inferred as `NDArrayFloat`

colour/characterisation/aces_it.py
```diff
@@ -1124,7 +1121,8 @@ def matrix_idt(
         XYZ_to_optimization_colour_model,
         finaliser_function,
     ) = optimisation_factory()
-    optimisation_settings = {
+
+    optimisation_settings: dict[str, Any] = {
```
jorenham, line 1125: without this, it'd be inferred as `dict[str, str]`, which due to its invariant type-parameters, would lead to it being rejected when passed as `**kwargs` to `minimize` below

colour/hints/__init__.py
```diff
@@ -149,18 +149,18 @@ def x(self) -> NDArray:  # noqa: D102
...
         ...
 
     @y.setter
-    def y(self, value: ArrayLike) -> None: ...
+    def y(self, value: ArrayLike, /) -> None: ...
 
     def __init__(self, *args: Any, **kwargs: Any) -> None: ...  # pragma: no cover
 
-    def __call__(self, x: ArrayLike) -> NDArray:  # noqa: D102
+    def __call__(self, x: NDArrayFloat) -> NDArray:  # noqa: D102
```
jorenham, line 163: input positions are contravariant, so assigning to this protocol, requires that the parameter types of that type are *less* strict than that of this `Protocol`. Put differently; narrowing the `ArrayLike` parameter type to `NDArrayFloat` results in a broader `Protocol` type.

colour/recovery/jakob2019.py
```diff
@@ -470,8 +464,8 @@ def find_coefficients_Jakob2019(
     )
 
     def optimize(
-        target_o: ArrayLike, coefficients_0_o: ArrayLike
-    ) -> Tuple[NDArrayFloat, float]:
+        target_o: NDArrayFloat, coefficients_0_o: NDArrayFloat
+    ) -> Tuple[NDArrayFloat, float | np.float64]:
```
jorenham, line 468: Both `result.fun: np.float64` and `errors.error: float` are assignable to `float | np.float64`, and on `numpy >= 2.2` it's even equivalent to `float64` (`float64` has been made a proper subtype of `builtins.float`)

colour/recovery/jakob2019.py
```diff
@@ -522,7 +516,7 @@ def optimize(
     if dimensionalise:
         coefficients = dimensionalise_coefficients(coefficients, cmfs.shape)
 
-    return coefficients, error
+    return coefficients, float(error)
```
jorenham, line 519: because `error: float | np.float64` at this point

colour/recovery/meng2015.py
```diff
@@ -177,12 +176,12 @@ def XYZ_to_sd_Meng2015(
 
     sd = sd_ones(cmfs.shape)
 
-    def objective_function(a: ArrayLike) -> NDArrayFloat:
+    def objective_function(a: NDArrayFloat) -> DTypeFloat:
         """Define the objective function."""
 
-        return cast(NDArrayFloat, np.sum(np.diff(a) ** 2))
+        return np.sum(np.square(np.diff(a)))
```
jorenham, line 182: In `numpy==2.2.0` there was a bug that caused `np.diff(a) ** 2` to be inferred as `np.signedinteger`. This has been resolved in `2.2.4`, but I figured that `square` would be a more stable approach.

jorenham:
Pardon the spam. I wanted to explain some of the non-obvious choices I made here. Feel free to resolve them as you see fit.

jorenham:
This is ready for review, in case you missed it :)

colour/io/luts/sony_spimtx.py
```diff
@@ -13,8 +13,7 @@
 import typing
 
 if typing.TYPE_CHECKING:
-    from pathlib import Path
-    from io import StringIO
+    from _typeshed import StrPath
```
KelSolaar, line 16: Sorry @jorenham for the late reply, it seems like my comment did not go through for some reason, I might have closed the laptop lid before they got pushed. At any rate, what I wrote was along the lines of I don't think we should rely on a type from a private import. In general, and unless there is a specific reason, we try to avoid doing this. I would prefer that we revert back to `str | Path` here.

KelSolaar (changes requested):
Thanks a lot! LGTM besides the private `_typeshed` import.

colour/io/luts/sony_spimtx.py
jorenham: `_typeshed` is bit of a special case that's maintained by Python itself:  https://github.com/python/typeshed/tree/main/stdlib/_typeshed

And although they don't officially guarantee permanent api stability, `StrPath` is one that's used *a lot*. For example, the Python standard library itself uses it 56 times: https://github.com/search?q=repo%3Apython%2Ftypeshed+StrPath+path%3A%2F%5Estdlib%5C%2F%2F&type=code, and libraries like NumPy also use it: https://github.com/search?q=repo%3Anumpy%2Fnumpy+StrPath&type=code

I know that the typeshed maintainers are also very careful about these sort of things. Any change they make will be ran through [mypy_primer](https://github.com/hauntsaninja/mypy_primer), and `colour` is listed in its projects. This means that if a typeshed change will result in a new typing error in `colour`, then it's likely that they will not go through with it.

KelSolaar: Ok, this is reasonable then. I suppose it is only available when type checking? If so, two paths:

1. Either we keep using `str | Path`
2. We alias it in `colour.hints.__init__` and use it across the board. We have `str | Path` / `Path | str` used in a few places.

jorenham:
```suggestion
    from _typeshed import StrPath
```

Yea; `_typeshed` is a type-checking only module, and type-checkers consider it as if it's part of the standard library. But since you're using `__future__.annotations`, that's not much of a problem. 

> 2. We alias it in `colour.hints.__init__` and use it across the board. We have `str | Path` / `Path | str` used in a few places.

I suppose that you could re-export it if you want to, but theoretically that's not much different from importing it directly 🤷🏻.

But either way, this is all pretty much out-of-scope for this PR, so option 1 it is

KelSolaar:
Thanks @jorenham, merged!

PR #1352, opened 2025-06-04
PR: Improve Docstrings

KelSolaar:
# Summary

This PR improves the docstrings by using Claude 4 via [Claude Code](https://www.anthropic.com/claude-code).

A few observations:

- This was expensive, more than anticipated: USD$265
   - Retrospectively, it would have been cheaper to get [Claude Max](https://www.anthropic.com/news/max-plan) for a month!
   - Cost was mostly driven by the fact I did a first test, which ( reverted because I did change our return value syntax, I deemed this undesirable .
   - The same prompt had to be executed multiple times because there are many instances where the guidelines were not fully respected
- Guidelines respect as is typically the case with a LLM is often miss or hit
- As I processed the docstrings sub-package by sub-package, the LLM took a slightly different approach every single time.

Was it worth it? I think so! Many small errors were fixed and the quality of the documentation has improved. However, for that particular task, I think it would be better to write a dedicated script that extracts the docstrings using the Python AST and process them by giving the object code and surrounding module code rather than having Claude Code doing the work itself. I was actually looking at whether we could implement a plugin on top of Ruff to do this or twist Pydocstyle's arm a bit!

The prompt I used is as follows:

---
**Task**: Update docstrings in all Python modules within the current sub-package recursively to improve wording and clarity.

**Processing Approach**:
- Think deeply
- Use external tools only for listing Python modules, not for docstring processing
- Process docstrings manually to ensure precision and compliance

**Mandatory Requirements**:

**Content Standards**:
- Apply technical writing style conforming to scientific research standards
- Use imperative mood (e.g., "Generate" not "Generates")
- Replace "given" with "specified" where contextually appropriate
- Make "luminance" and "lightness" lowercase unless sentence-initial or after colon
- Preserve all existing docstring examples without modification
- Add "Raises" section when exceptions are raised by the function
- For docstrings beginning with "Return", consider contextually appropriate alternatives
- Focus on clarity and precision in scientific context
- Preserve all Sphinx markers (:math:, :param:, etc.) without modification
- Maintain all reStructuredText emphasis markers around words, e.g., * in *luminance*, `` in ``RGB``, etc...
- Preserve reStructuredText section headers exactly as written
- Keep empty line between docstring and code

**Restrictions**:
- Do NOT add parameter types (already documented via signature types)
- Do NOT modify any object code outside docstrings
- Do NOT edit module headers unless they contain clear errors
- Do NOT process unit test modules
- Do NOT change "colour" to "color" or "colourspace" to "colorspace"/"color space"
- Do NOT change any reStructuredText section headers, i.e., all the headers underlined with - or = must be kept as is

**Formatting**:
- Wrap all lines at exactly 79 characters

**Verification Checklist**:
Before completing each docstring update, verify:
- [ ] All Sphinx markers preserved
- [ ] Emphasis markers preserved
- [ ] Section headers preserved
- [ ] Imperative mood used consistently
- [ ] "Raises" section added if exceptions present
- [ ] No code modifications outside docstrings
- [ ] British spelling preserved
- [ ] 79-character line wrapping maintained
---

# Preflight

**Code Style and Quality**

**Documentation**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 *.pyo
 .DS_Store
 .coverage*
+CLAUDE.md
 uv.lock
 
 # Common Directories
--- a/colour/adaptation/__init__.py
+++ b/colour/adaptation/__init__.py
@@ -159,7 +159,7 @@ def chromatic_adaptation(
     **kwargs: Any,
 ) -> NDArrayFloat:
     """
-    Adapt given stimulus from test viewing conditions to reference viewing
+    Adapt the specified stimulus from test viewing conditions to reference viewing
     conditions.
 
     Parameters
@@ -187,8 +187,8 @@ def chromatic_adaptation(
         Noise component in fundamental primary system.
     Y_o
         {:func:`colour.adaptation.chromatic_adaptation_CIE1994`},
-        Luminance factor :math:`Y_o` of achromatic background normalised to
-        domain [0.18, 1] in **'Reference'** domain-range scale.
+        Luminance factor :math:`Y_o` of achromatic background normalised
+        to domain [0.18, 1] in **'Reference'** domain-range scale.
     direction
         {:func:`colour.adaptation.chromatic_adaptation_CMCCAT2000`},
         Chromatic adaptation direction.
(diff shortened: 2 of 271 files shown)


nick-shaw:
Sometimes the global change from "given" to "specified" makes things read a little oddly, e.g.
```
-        If the given specification doesn't exist in *Munsell Renotation System*
+        If the specified specification doesn't exist in *Munsell Renotation System*
```

KelSolaar:
> Sometimes the global change from "given" to "specified" makes things read a little oddly, e.g.
> 
> ```
> -        If the given specification doesn't exist in *Munsell Renotation System*
> +        If the specified specification doesn't exist in *Munsell Renotation System*
> ```

Good catch, fixed!

PR #1358, opened 2025-08-24
faster type-checking by reducing `np.array` expressions

jorenham:
On my laptop this reduces the pyright runtime by 10 seconds (~15% faster), and the mypy runtime by ~1 minute (~30% faster).


jorenham:
> Thanks @jorenham! I think we could apply this in other places of the codebase. I will take a look.

Nice! In case you weren't aware yet; you can see the per-file analysis times by running `pyright --stats --verbose`, so that you can see where the bottlenecks are.

PR #1408, opened 2026-06-06
PR: Add TLCI-2012 and TLMF-2013 quality metrics

### LINKED ISSUE #974: [FEATURE]: Implement TLCI / TLMF calculation
### Description

TLCI-2012 is a similar measurement to CRI, with different swatches and an expanded algorithm.
TLMF-2013 is an additional algorithm that matches a the luminaire, given camera data. This one may be harder, as the user would need to supply this data.

The brief is here:
https://tech.ebu.ch/tlci-2012

The software zip has a little bit of information, which expanded, along with all of the math, in the tech brief here:

https://tech.ebu.ch/docs/tech/tech3355.pdf

cmuellner:
## Summary

Add EBU Tech 3355 TLCI-2012 and TLMF-2013 quality metrics.

This implements:

- Television Lighting Consistency Index, TLCI-2012.
- Television Luminaire Matching Factor, TLMF-2013.
- EBU Tech 3355 camera, display, saturation, and reference-selection data.
- EBU Tech 3355 Appendix 2 Planckian and daylight chromaticity tables used
  for TLCI CCT selection.
- EBU Tech 3355 Appendix 3 daylight radiation vector data used for TLCI
  daylight references.
- EBU Tech 3355 Appendix 4 test colour sample data used by TLCI/TLMF,
  including the documented BBC R&D Report 1988/2 and unpublished BBC R&D
  provenance.
- Public `colour` and `colour.quality` API exports.
- Documentation, examples, and bibliography entries.

Fixes #974.

## Implementation Notes

TLCI/TLMF was not a straightforward transcription from the public
specification. A few parts of EBU Tech 3355 require careful interpretation:

- The CCT selection is defined by Tech 3355, not by Colour's existing
  CCT/D_uv methods such as Ohno (2013) or Robertson (1968). Tech 3355 uses
  the Appendix 2 Planckian and daylight chromaticity tables, scans adjacent
  locus points, and computes the normal intersection using the section 1.1.1
  slope/angle procedure.
- The Planckian and daylight CCT loci are the Appendix 2 lookup tables, not
  regenerated blackbody or CIE daylight loci. The Planckian and daylight loci
  also remain separate through the 3400-5000 K mixed-reference region.
- The Planckian reference spectrum is Tech 3355-specific. Section 1.1.2.1
  equation [9] uses a simplified Planckian expression with wavelength in
  nanometres, a `1.435e7 nm K` radiation constant, and normalisation at
  560 nm; this is not delegated to Colour's generic blackbody helper.
- The daylight reference spectrum is also Tech 3355-specific. Section 1.1.2.2
  uses Appendix 3 daylight radiation vectors and Tech 3355 coefficients for
  `M`, `M1`, and `M2`; this differs from Colour's general
  `sd_CIE_illuminant_D_series` helper, which uses the library CIE D-series
  basis data and its standard coefficient/rounding path.
- The 3400-5000 K mixed-reference spectrum is interpolated from fixed
  Planckian 3400 K and daylight 5000 K endpoint spectra, after normalising
  each endpoint at 560 nm. It is not generated from Planckian/daylight spectra
  at the test source CCT.
- The Appendix 4 test-colour samples are the mandatory TLCI/TLMF data, but
  they are not simply a direct copy of the published BBC R&D Report 1988/2
  ColorChecker table. Some coloured samples differ slightly from the published
  BBC table, and the greyscale samples are attributed by Tech 3355 to
  unpublished BBC R&D data. The dataset is therefore documented as Tech 3355
  data, with BBC R&D Report 1988/2 as partial provenance.
- The Tech 3355 `d_uv` calculation differs from Colour's existing
  CCT/D_uv methods such as Ohno (2013) and Robertson (1968): it uses the
  Tech 3355 reference loci, normalises the distance by 0.0054, and reports
  positive values on the magenta side and negative values on the green side.
  This follows the section 1.1.1 prose that reverses the sign for green-side
  offsets, where `u_T < u_L`, and the section 1.1.2.3 equations [16]-[17]
  labels (`d > 0` towards magenta, `d <= 0` towards green). The EBU reference
  binary appears to report this secondary `d_uv` value with the opposite sign,
  but the implementation follows the published specification. The reported
  `d_uv` also includes the Tech 3355 mixed-region adjustment between 4000 K
  and 5000 K, where the Planckian and daylight loci are not joined.
- TLMF shares the TLCI camera/display pipeline but uses the reference luminaire
  balance with a test-luminaire exposure correction, matching the TLMF
  procedure instead of independently white-balancing both sources as TLCI does.
  The exposure correction uses the reference/test neutral luma ratio after the
  camera matrix.
- TLCI-2012 aggregates colour differences over only the first 18 coloured
  ColorChecker samples, while TLMF-2013 uses all 24 samples, including the
  greyscale patches.
- The camera BT.709 signal path preserves the studio-swing headroom described
  by Tech 3355 section 2, equation [61], capping at code value 255 relative
  to the nominal 16-235 range instead of clipping at nominal white.
- The Tech 3355 colour-difference calculation follows the section 1.5
  CIEDE2000 path with unity `k` factors, using Colour's existing CIE 2000
  implementation for the standard colour-difference calculation.

## Validation

EBU Tech 3355 does not distribute official reference test vectors. The tests
therefore use reproducible generated spectral distributions covering the main
specified algorithm paths:

- Planckian sources.
- Tech 3355 daylight-reference sources.
- Mixed-reference spectra in the 3400-5000 K transition.
- Reference-locus distance sign checks for green/magenta chromaticity
  offsets.
- Phosphor-like LED spectra.
- RGB LED spectra.
- Public in-tree CIE/NIST LED spectral distributions as regression/stress
  coverage.
- TLMF identical, close-match, mismatch, and mixed-reference pairs.

Additional local validation sets, including a 1000-case TLCI and 1000-pair
TLMF diagnostic set, were used to stress the CCT branches, mixed-reference
transition, green/magenta `d_uv` sign, LED spectra, fixture-like spectra, and
low-score TLMF cases. The committed tests use only deterministic,
redistributable spectra and public in-tree spectral data; they do not depend
on local measurement files.

Because no official reference vectors are published, this PR cannot claim
independent conformance from official test data. The validation strategy is
instead:

- implement each calculation step from the published specification;
- reuse existing Colour mathematics for spectral integration, colour
  transfer functions, and colourspace conversions;
- keep TLCI-specific code only where Tech 3355 defines a calculation that is
  not covered by an existing Colour routine;
- commit a smaller redistributable validation set that exercises the same
  algorithm branches.

## Tests

- Relevant pre-commit hook tools run directly: `isort`, `ruff format`,
  `ruff check`, `codespell`, `blacken-docs`, `check-added-large-files`,
  `check-case-conflict`, `check-merge-conflict`, `check-symlinks`,
  `check-yaml`, `debug-statement-hook`, `end-of-file-fixer`,
  `mixed-line-ending`, `name-tests-test`, and `trailing-whitespace-fixer`.
- `prettier` and the RST pygrep hooks were run through `prek`; no
  `requirements.txt` files are touched by the commit.
- `uv run ruff check colour/quality/tlci.py colour/quality/datasets/tlci2012.py colour/quality/datasets/__init__.py colour/quality/tests/test_tlci.py`
- `uv run pyright --threads --skipunannotated colour/quality/tlci.py colour/quality/datasets/tlci2012.py colour/quality/tests/test_tlci.py`
- `uv run python -m doctest colour/quality/tlci.py`
- `uv run pytest --doctest-modules colour/quality/tlci.py`
- `uv run pytest colour/quality/tests/test__init__.py colour/quality/tests/test_tlci.py`

## Preflight

**Code Style and Quality**

- [n/a] New transformations have been added to the _Automatic Colour Conversion Graph_.
- [n/a] New transformations have been exported to the relevant namespaces, e.g. `colour`, `colour.models`.

**Documentation**


cmuellner:
CI is currently failing for two reasons.

The GitHub Actions failure appears unrelated to this PR. The failing job is `windows-latest - Python 3.12`, and the failures are all urllib.error.HTTPError: HTTP Error 429: Too Many Requests in existing network-dependent tests while downloading Munsell ONNX assets / test resources via download_url(). The new TLCI/TLMF tests are not among the failures, and the other OS/Python jobs passed, so this looks like external rate limiting rather than a behavioral regression.

Codacy reports 65 new issues, mostly in the new TLCI/TLMF files:
- 24x Bandit_B101 for pytest assert usage in colour/quality/tests/test_tlci.py.
- 41x Prospector/pydocstyle findings, mostly docstring spacing/formatting rules
  such as D202, D200, D203, D205, D212, D413, and D415.
- Codacy also reports one new duplication item in
  colour/quality/datasets/tlci2012.py, which appears to be related to reference
  dataset content.

My read is that many of these findings conflict with the existing project style: the test suite widely uses pytest assert, Ruff ignores S101 for test_* files, and several of the pydocstyle rules reported by Codacy are explicitly ignored by
the local Ruff configuration or conflict with existing module/class docstring patterns. That said, if maintainers prefer satisfying Codacy on this PR, I can adjust the new files accordingly or add targeted suppressions/configuration. I’m
fine with either direction; the main tradeoff is consistency with existing style versus making Codacy green without changing Codacy configuration.

thomasmansencal:
Hi @cmuellner,

Thanks a lot! Great work. #1406 is pending and is incompatible with it at the moment so, the best way would be for me to merge it and rebase #1406 on top of it and bring it up to Array API standard compatibility.

Don't worry about Codacy and the ONNX download CI issue.

cmuellner:
Thanks! If you would prefer the opposite sequencing for any reason (Python Array API lands first), I’m also happy to adjust this PR accordingly.

thomasmansencal:
The issue is that we might change some of our Array API Standard implementation. I will keep you posted!

colour/quality/tlci.py
```diff
@@ -0,0 +1,766 @@
...
+        values indicate that the test source is towards green
+        (:math:`u_T < u_L` in CIE 1960 UCS); positive values indicate that it
+        is towards magenta. Absolute values above 1.0 indicate reduced CCT
+        reliability.
+    """
+
+    name: str
+    tlci: float
+    delta_E_a: float
+    delta_E: NDArrayFloat
```
thomasmansencal: ### Minor

Could we rename it to `delta_E_s`, the way `cfi2017` does, i.e., the per-sample colour differences pairing with the existing `delta_E_a`? Same field on the TLMF specification at line 126.

colour/quality/tlci.py
```diff
@@ -0,0 +1,766 @@
...
+__all__ = [
+    "TLCI2012_Specification",
+    "TLMF2013_Specification",
+    "television_lighting_consistency_index",
+    "television_luminaire_matching_factor",
+]
+
+
+@dataclass()
+class TLCI2012_Specification:
```
thomasmansencal: ### Major

For consistency with the sibling specifications, could we follow the `<Category>_Specification_<Method>` pattern (`ColourRendering_Specification_CRI`, `ColourQuality_Specification_ANSIIESTM3018`)? So `ColourQuality_Specification_TLCI2012` / `ColourQuality_Specification_TLMF2013`.

colour/quality/tlci.py
```diff
@@ -0,0 +1,766 @@
...
+    d_uv
+        Distance from the reference locus in TLCI ``d`` units. Negative
+        values indicate that the test source is towards green
+        (:math:`u_T < u_L` in CIE 1960 UCS); positive values indicate that it
+        is towards magenta. Absolute values above 1.0 indicate reduced CCT
+        reliability.
+    """
+
+    name: str
+    tlci: float
```
thomasmansencal: ### Major

Could we name the score field `Q_a`, to match the other quality specifications (`cri`, `cqs`)? `_Q_from_delta_E` already returns `Q_a`, so it lines up nicely.

colour/quality/tlci.py
```diff
@@ -0,0 +1,766 @@
...
+    delta_E_a: float
+    delta_E: NDArrayFloat
+    CCT: float
+    d_uv: float
```
thomasmansencal: ### Minor

Small consistency one: `d_uv` -> `D_uv`, matching `cfi2017`. The docstring already explains the EBU `d` mapping; the only thing to add there is the sign note (see the comment on the sign below).

colour/quality/tlci.py
```diff
@@ -0,0 +1,766 @@
...
+    from colour.hints import Literal, NDArrayBoolean, NDArrayFloat
+
+__author__ = "Colour Developers"
+__copyright__ = "Copyright 2013 Colour Developers"
+__license__ = "BSD-3-Clause - https://opensource.org/licenses/BSD-3-Clause"
+__maintainer__ = "Colour Developers"
+__email__ = "colour-developers@colour-science.org"
+__status__ = "Production"
+
+__all__ = [
```
thomasmansencal, line 64: ### Major

This module leans heavily on private helpers (17 `def _*`), whereas the siblings (`cfi2017`, `cri`, `cqs`, `ssi`, `tm3018`) expose their computation helpers publicly via `__all__`. Once the trivial ones are inlined, it would be great to promote the substantive helpers (`sd_planckian_TLCI2012`, `sd_daylight_TLCI2012`, `sd_reference_illuminant_TLCI2012`, `uv_to_CCT_TLCI2012`) to public, prefix-named, and list them here, so it matches the family.

colour/quality/tlci.py
```diff
@@ -0,0 +1,766 @@
...
+# EBU Tech 3355 section 1.4.1, equation [28].
+_DISPLAY_GAMMA = 2.4
+# EBU Tech 3355 section 2, equation [61], maps the notional camera signal
+# R_C' to television coding as 16 + 219 R_C'. The text states that signals
+# need not be clipped at nominal peak level 235; the corresponding full-scale
+# signal at code value 255 is therefore:
+_STUDIO_SWING_WHITE = (255 - 16) / (235 - 16)
+
+
+def _align_sd(sd: SpectralDistribution, shape: SpectralShape) -> SpectralDistribution:
```
thomasmansencal: ### Minor

A couple of these are thin wrappers we could inline to keep things simple: `_align_sd` / `_align_msds` just pass `"Align", copy=False` to `reshape_sd` / `reshape_msds`, and `_sd_to_XYZ` / `_sd_to_uv` are each single-use. `_normalise_sd_560` does real work, so that one is fine to keep.

colour/quality/tlci.py
```diff
@@ -0,0 +1,766 @@
...
+    _, CCT, uv_locus, is_daylight = min(candidates, key=lambda candidate: candidate[0])
+
+    return (
+        CCT,
+        uv_locus,
+        is_daylight,
+    )
+
+
+def _normalise_sd_560(sd: SpectralDistribution) -> SpectralDistribution:
```
thomasmansencal: ### Minor

Small naming consistency: the SD-returning helpers mostly use the `sd_*` prefix (`_sd_planckian_TLCI2012`, `_sd_daylight_TLCI2012`), but `_normalise_sd_560` uses a suffix even though it also returns an SD. The ecosystem convention is the `sd_*` prefix (`sd_blackbody`, `sd_gaussian`); could we align it?

colour/quality/tlci.py
```diff
@@ -0,0 +1,766 @@
...
+        )
+
+    # EBU Tech 3355 section 1.1.1, equation [8], normalised to 0.0054 per
+    # the prose following that equation.
+    # Tech 3355 section 1.1.1 reverses the sign for green-side offsets,
+    # where uT < uL. Section 1.1.2.3 equations [16]-[17] label d > 0 as
+    # magenta and d <= 0 as green.
+    d_uv = as_float_scalar(np.linalg.norm(uv - uv_locus) / 0.0054)
+
+    if uv[0] < uv_locus[0]:
```
thomasmansencal, line 569: ### Question

The `d` sign keys off `u`, which is exactly right per Tech 3355 section 1.1.1 (green side -> u_T < u_L -> reversed), so no bug here. One thing worth knowing is that this is the opposite sign to the EBU TLCI application (its manual, section 2.1.3.1.2, shows d > +1/2 in green and d < -1/2 in magenta). Since many people compare against that tool, should we match the standard (as you have) or the application? Either way, it would help to note the convention and the disagreement in the `D_uv` docstring.

colour/quality/tlci.py
```diff
@@ -0,0 +1,766 @@
...
+    shape: SpectralShape,
+) -> NDArrayFloat:
+    """
+    Compute raw camera RGB signals for ``patch_names`` under ``sd_illuminant``.
+    """
+
+    illuminant = _align_sd(sd_illuminant, shape)
+    camera = _align_msds(msds_camera, shape)
+
+    signals = np.empty((len(patch_names), 3))
```
thomasmansencal: ### Minor

This loops over the 24 patches calling `sd_to_XYZ` each time, and runs twice per metric. We could collapse it to a single `msds_to_XYZ(msds, camera, illuminant, method="Integration")`.

colour/quality/tlci.py
```diff
@@ -0,0 +1,766 @@
...
+
+    return (
+        as_float_scalar(distances[index]),
+        as_float_scalar(CCT),
+        uv_intersections[index],
+        is_daylight,
+    )
+
+
+def _nearest_locus_sample_TLCI2012(
```
thomasmansencal, line 264: ### Minor

A few paths look untested: the `_nearest_locus_sample_TLCI2012` fallback, the all-excluded `raise` in `_Q_from_delta_E`, and the non-default `camera` argument.

colour/quality/datasets/tlci2012.py
```diff
@@ -0,0 +1,2633 @@
...
+
+Columns are :math:`\\bar{r}`, :math:`\\bar{g}`, :math:`\\bar{b}` from
+380 nm to 760 nm at 5 nm intervals. Each column sums to 1.000000.
+
+References
+----------
+:cite:`EuropeanBroadcastingUnion2017`
+"""
+
+MSDS_CAMERA_CURVES_TLCI2012: MultiSpectralDistributions = (
```
thomasmansencal: ### Major

Please use `colour.characterisation.RGB_CameraSensitivities` here, which is built for that purpose, rather than a plain `MultiSpectralDistributions`. It also resolves the `_SENSITIVITIES_` vs `_CURVES_` naming nicely, since the data is camera sensitivities.

`MSDS_CAMERA_CURVES_TLCI2012` --> `MSDS_CAMERA_SENSITIVITIES_TLCI2012`

colour/quality/datasets/tlci2012.py
```diff
@@ -0,0 +1,2633 @@
...
+*EBU Tech 3355* Appendix 2 daylight radiator chromaticities.
+
+Columns are CCT, :math:`x`, and :math:`y`.
+
+References
+----------
+:cite:`EuropeanBroadcastingUnion2017`
+"""
+# fmt: off
+DATA_DAYLIGHT_BASIS_TLCI2012: NDArrayFloat = np.array(
```
thomasmansencal, line 528: ### Note

Worth a short `Notes` block here: Tech 3355 attributes these vectors to CIE 15:2004, but the 5 nm values differ from CIE Colorimetry (4th Ed) Table 6, 37 of the 77 S0 samples differ, all at the odd 5 nm wavelengths (e.g. S0@385 = 62.45 vs 64.60), while the 10 nm Judd anchors match.

colour/quality/tests/test_tlci.py
```diff
@@ -0,0 +1,378 @@
...
+
+    return normalised_sd(
+        0.72 * gaussian(wavelengths, 455, 14)
+        + 1.00 * gaussian(wavelengths, 535, 18)
+        + 0.86 * gaussian(wavelengths, 625, 18),
+        "rgb-led-balanced",
+    )
+
+
+def test_tlci_datasets() -> None:
```
thomasmansencal: ### Minor

For consistency with the other quality tests, could we wrap these in one class per public function, with a single `test_<function>` method bundling the cases (see `TestColourRenderingIndex`, `TestColourFidelityIndexCIE2017`, which also adds a class per public helper)? So `TestTelevisionLightingConsistencyIndex` / `TestTelevisionLuminaireMatchingFactor`, plus a class per helper once they are public, ordered to match the source.

colour/quality/datasets/tlci2012.py
```diff
@@ -0,0 +1,2633 @@
...
+)
+"""
+*EBU Tech 3355* Appendix 4 test colour sample spectral distributions.
+
+References
+----------
+:cite:`EuropeanBroadcastingUnion2017`, :cite:`BBCResearchDepartment1988`
+"""
+
+DATA_CAMERA_SENSITIVITIES_TLCI2012: NDArrayFloat = DATA_CAMERA_CURVES_TLCI2012
```
thomasmansencal: ### Minor

Do we need `DATA_CAMERA_CURVES_TLCI2012` at all given my previous comment? What was the rationale?

colour/quality/tlci.py
```diff
@@ -0,0 +1,766 @@
...
+    Examples
+    --------
+    >>> from colour import SDS_ILLUMINANTS
+    >>> television_lighting_consistency_index(  # doctest: +ELLIPSIS
+    ...     SDS_ILLUMINANTS["FL2"]
+    ... )
+    np.float64(29.49254175...)
+    """
+
+    camera_name = camera or "EBU Standard Camera"
```
thomasmansencal: ### Minor

`camera or "EBU Standard Camera"` could be `optional(camera, "EBU Standard Camera")`; it also avoids `or` swallowing an empty string.

colour/quality/tlci.py
```diff
@@ -0,0 +1,766 @@
...
+    """
+    Return the *EBU Tech 3355* Planckian reference spectrum for ``CCT``.
+    """
+
+    # EBU Tech 3355 section 1.1.2.1, equation [9], uses a simplified
+    # Planckian expression with wavelength in nanometres, a 1.435e7 nm K
+    # radiation constant, and normalisation at 560 nm. The implementation
+    # returns unity at 560 nm instead of 100; the global scale is absorbed by
+    # the camera white-balance step.
+    wavelengths = np.asarray(shape.wavelengths)
```
thomasmansencal: ### Pedantry

`np.asarray(shape.wavelengths)` is redundant, `shape.wavelengths` is already an array (same in the tests).

colour/quality/tlci.py
```diff
@@ -0,0 +1,766 @@
...
+        return sd_to_XYZ(sd, cmfs, method="Integration")
+
+
+def _sd_to_uv(sd: SpectralDistribution, shape: SpectralShape) -> NDArrayFloat:
+    """Compute CIE 1960 UCS *uv* chromaticity coordinates for ``sd``."""
+
+    return UCS_to_uv(XYZ_to_UCS(_sd_to_XYZ(sd, shape)))
+
+
+@functools.cache
```
thomasmansencal: ### Minor

`@functools.cache` on a no-argument constant builder could just be a module-level constant.

colour/quality/datasets/__init__.py
```diff
@@ -1,9 +1,43 @@
...
+    "DATA_DAYLIGHT_LOCUS_TLCI2012",
+    "DATA_PLANCKIAN_LOCUS_TLCI2012",
+    "DATA_TCS_TLCI2012",
+    "MATRIX_TLCI2012_CAMERA",
+    "MATRIX_TLCI2012_DISPLAY",
+    "MATRIX_TLCI2012_SATURATION",
+    "MSDS_CAMERA_CURVES_TLCI2012",
+    "MSDS_DAYLIGHT_BASIS_TLCI2012",
+    "SDS_TCS_TLCI2012",
+    "MSDS_CAMERA_SENSITIVITIES_TLCI2012",
```
thomasmansencal, line 31: ### Pedantry

Small ordering thing: `MSDS_CAMERA_SENSITIVITIES_TLCI2012` sits after `SDS_TCS_TLCI2012` in `__all__`. The import block above is sorted, so it would be nice to sort `__all__` to match.

colour/quality/tests/test_tlci.py
```diff
@@ -0,0 +1,378 @@
...
+SPECTRAL_SHAPE_TLCI_TLMF_TESTS = SpectralShape(380, 760, 5)
+
+
+def gaussian(wavelengths: np.ndarray, centre: float, width: float) -> np.ndarray:
+    """Return a unit Gaussian spectral lobe."""
+
+    return np.exp(-0.5 * ((wavelengths - centre) / width) ** 2)
+
+
+def normalised_sd(values: np.ndarray, name: str) -> SpectralDistribution:
```
thomasmansencal: ### Important

These free helpers use the `_sd` suffix (`normalised_sd`, `planckian_sd`); the ecosystem convention is the `sd_*` prefix, if you would like to align them.

thomasmansencal (changes requested):
Thanks @cmuellner!

I left a few comments, most of them being about consistency with the surrounding code. When they are addressed, I will merge, do a polish pass on everything I might have missed and rebase the Array API PR on top of develop.

cmuellner:
Thanks for the thorough review @thomasmansencal, and apologies for the slow turnaround.

All points are addressed and the branch is rebased onto current `develop`.

**Naming / API**

- `TLCI2012_Specification` / `TLMF2013_Specification` → `ColourQuality_Specification_TLCI2012` / `ColourQuality_Specification_TLMF2013`, propagated to `colour/__init__.py`, `colour/quality/__init__.py` and `docs/colour.quality.rst`.
- Score field → `Q_a`, per-sample differences → `delta_E_s`, `d_uv` → `D_uv`.
- Promoted `sd_planckian_TLCI2012`, `sd_daylight_TLCI2012`, `sd_reference_illuminant_TLCI2012` and `uv_to_CCT_TLCI2012` to public and listed them in `__all__`. Private helpers are down from 17 to 10. I followed the `cfi2017` pattern: the helpers are listed in the module `__all__`, while `colour.quality` re-exports only the specification and the top-level function.
- `MSDS_CAMERA_CURVES_TLCI2012` → `MSDS_CAMERA_SENSITIVITIES_TLCI2012`, now built as `RGB_CameraSensitivities`. `DATA_CAMERA_CURVES_TLCI2012` is gone.

**Simplification**

- Inlined `_align_sd`, `_align_msds`, `_sd_to_XYZ` and `_sd_to_uv`; `_normalise_sd_560` kept and renamed `_sd_normalise_560_TLCI2012`.
- Collapsed the per-patch integration loop to a single `msds_to_XYZ(..., method="Integration")`.
- Replaced `@functools.cache` with a module-level constant, `camera or ...` with `optional(...)`, and dropped the redundant `np.asarray(shape.wavelengths)`.

**Tests and docs**

- Restructured into one class per public function and per public helper, ordered to match the source, and renamed the module-level fixtures to the `sd_*` prefix.
- Added coverage for the three untested paths: the `_nearest_locus_sample_TLCI2012` fallback, the all-excluded `raise` in `_Q_from_delta_E`, and the non-default `camera` argument.
- Added the `Notes` block on the Appendix 3 daylight radiation vectors documenting the deviation from *CIE Colorimetry* Table 6.
- Sorted `__all__` in `colour/quality/datasets/__init__.py` to match the import block.

Locally: `ruff` and `ruff format` clean, `pyright --threads --skipunannotated` reports 0 errors, and the `colour/quality` and `colour/tests/test__all__.py` suites pass.

One item is still an open decision rather than a change: the **`D_uv` sign convention** (your question on `tlci.py:417`). I have kept the published *Tech 3355* convention and documented the disagreement with the EBU application, but I am happy to flip it if you prefer matching the tool — details in that thread.

Also, is the sequencing from June still what you want, i.e. this lands first and #1406 rebases on top? Happy to work the other way round if the Array API work has moved on since.

colour/quality/tlci.py
cmuellner: Done, renamed on both the TLCI and TLMF specifications.

colour/quality/tlci.py
cmuellner: Done. Renamed to `ColourQuality_Specification_TLCI2012` and `ColourQuality_Specification_TLMF2013`, and propagated to `colour/__init__.py`, `colour/quality/__init__.py` and `docs/colour.quality.rst`.

colour/quality/tlci.py
cmuellner: Done.

colour/quality/tlci.py
cmuellner: Done. The sign note is now a `.. note::` block in the `D_uv` field docstring; see my reply on the sign convention thread for the wording and the open question.

colour/quality/tlci.py
cmuellner: Done.

colour/quality/tlci.py
cmuellner: Done, it is now `_sd_normalise_560_TLCI2012`.

colour/quality/tlci.py
cmuellner: Thanks, this is the one item that is hard to decide.

I have kept the published standard, so the current behaviour is unchanged: green gives a negative `D_uv`, magenta positive, per *Tech 3355* section 1.1.1 and equations [16]-[17].

I have documented the disagreement in the `D_uv` field docstring:
> This sign convention follows the published *EBU Tech 3355* specification (section 1.1.1 and equations [16]-[17]). The *EBU TLCI* application reports the opposite sign: its manual, section 2.1.3.1.2, shows :math:`d > +1/2` towards green and :math:`d < -1/2` towards magenta. Values compared against that tool will therefore appear negated.

My reasoning for keeping the standard is that it is the normative document, and that the negation is trivial for anyone cross-checking against the application as long as it is documented. The counter-argument is real though: in practice most people meet TLCI through the EBU tool, and a silent sign flip against it is a plausible source of confusion.

Both are one small change apart. Flipping means negating the sign at the single computation site and updating the docstring plus the `test_sd_reference_illuminant_TLCI2012_D_uv_sign` expectations. Let me know which you prefer and I will make it so; there is also a middle option of keeping the standard and adding the application convention as a note in the public function docstrings, if you want it more prominent than the field docstring.

colour/quality/tlci.py
cmuellner: Done.

colour/quality/tlci.py
cmuellner: Done, all three are covered now.

colour/quality/datasets/tlci2012.py
cmuellner: Done.

colour/quality/datasets/tlci2012.py
cmuellner: Done.

colour/quality/tests/test_tlci.py
cmuellner: Done.

colour/quality/datasets/tlci2012.py
cmuellner: You are right that it was redundant, it only existed as the raw array behind the `MultiSpectralDistributions`. It is removed, along with its export.

colour/quality/tlci.py
cmuellner: Done.

colour/quality/tlci.py
cmuellner: Done.

colour/quality/tlci.py
cmuellner: Done.

colour/quality/datasets/__init__.py
cmuellner: Done.

colour/quality/tests/test_tlci.py
cmuellner: Done.

colour/quality/tlci.py
cmuellner: Done.

thomasmansencal:
Thanks a lot @cmuellner, I will merge this one :)

PR #1419, opened 2026-09-09
Array api improvements to preserve computaional graph

MichaelMauderer:
# Summary

# Preflight

**Code Style and Quality**

- [ ] Unit tests have been implemented and passed.
- [ ] Pre-commit hooks have been run and passed.
- [ ] New transformations have been added to the _Automatic Colour Conversion Graph_.
- [ ] New transformations have been exported to the relevant namespaces, e.g. `colour`, `colour.models`.

**Documentation**

- [ ] New features are documented along with examples if relevant.


colour/utilities/array_api/interpolation/__init__.py
```diff
@@ -0,0 +1,52 @@
+"""
+Array API Interpolation
```
thomasmansencal, line 2: Wondering if we should move all this in `colour.algebra` and make a proper `interpolation/` sub-package that contains all this.

MichaelMauderer: We could do that. My thinking here was that this something so generic as not needing to be `colour` at all in a sense. if it wouldn't add the extra overhead Iw ould actually want to put it intoa a separate `array-api-extension`/`scipy-array-api` package.

